PR #573, opened 2023-02-27
[DO NOT MERGE] feat: rework parity signer onboarding

0xKheops:


alecdwm (commented):
IMO we actually shouldn't merge this - the existing UI isn't great (we need to fix it!) but this solution we came up with doesn't work *at all* with parity vault 😢

Wdyt?

0xKheops:
obsolete

PR #856, opened 2023-06-17
feat: ✨ automatic gasLimit boost

### LINKED ISSUE #753: Estimated gas to low
image

This happens quite rarely, and somehow disappears after few minutes, like if the chain was self-healing.
The estimated gas returned by the chain is a little to low, when doing the validation estimateGas call (with from filled), it throws an error.
Going to custom gas settings and saving (without changing gasLimit) solves the issue.

Maybe we could add 1% to the estimateGas when setting gasLimit ?
it would make our "Estimated Fee" and "MaxFee" higher than it should tho.

Note : screenshot at block 3556467

0xKheops:
- [x] Fixes https://github.com/TalismanSociety/talisman/issues/753

Hard to validate the fix because it doesn't happen often and RPC auto corrects the gas estimation after few minutes when it does.
Will keep this PR as draft for a week and do all my DPS runs with it until i can confirm it fixes the issue

### DIFF
--- a/apps/extension/src/core/domains/ethereum/helpers.ts
+++ b/apps/extension/src/core/domains/ethereum/helpers.ts
@@ -122,7 +122,8 @@ const TX_GAS_LIMIT_MIN = BigNumber.from("21000")
 export const getGasLimit = (
   blockGasLimit: BigNumberish,
   estimatedGas: BigNumberish,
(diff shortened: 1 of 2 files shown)


0xKheops:
closing - this will be fixed in #1007 instead

PR #866, opened 2023-06-22
feat: ✨ batch import accounts

### LINKED ISSUE #865: Batch Import P.JS

0xKheops:
- Closes #865 

Progress : 
- [x] Account picker
- [x] Unlock multiple accounts at once
- [x] Import single or multi
- [x] Prevent importing external/hardware accounts (cuz we don't know the origin)
- [x] New file picker
- [x] Display balances in account picker
- [x] Live update progress bar when trying passwords
- [x] QA
- [x] Cleanup


apps/extension/src/ui/apps/dashboard/routes/AccountAddJson/UnlockJsonAccountsButton.tsx
```diff
@@ -0,0 +1,158 @@
...
+              className="bg-grey-700 absolute left-0 top-0 h-5 w-full rounded-lg transition-transform ease-out"
+              style={unlockAttemptProgressStyle}
+            ></div>
+            <div
+              className="bg-primary-500 absolute left-0 top-0 h-5 w-full rounded-lg transition-transform duration-300 ease-out"
+              style={progressStyle}
+            ></div>
+          </div>
+          <div className="text-body-secondary my-16">
+            {t("Enter passwords until all selected accounts are unlocked")}
```
chidg: Maybe `Enter the passwords for each of your selected accounts, until all accounts unlocked`.

chidg (changes requested):
Great code and UI
Few small points
- I'm not 100% happy with the 'input password for multiple accounts' modal UX. It doesn't feel totally clear what's going on. I suggested a minor copy change, and I think we should release and get user feedback. When we initially discussed this I imagined a UI where the password input took place in the context of the account list, so that it was more clear that the passwords were for the individual accounts. Anyway, let's release this and see
- Some validation on JSON upload would be good. Can we validate that the shape of the json data is correct? Currently you can load any random json file and the page will not display any error message.
- When a file is loaded where all of its accounts are already imported, maybe we should have a special message like "No importable accounts present"
Screenshot 2023-07-07 at 3 36 02 pm
- Can we have a tooltip on the locked/unlocked icon in the accounts list - I wasn't totally clear what that meant the first time I tried.
Screenshot 2023-07-07 at 11 24 43 am

0xKheops:
Implemented an alert like this if all importable accounts are already in the keyring

image

chidg (approved):
LGTM!

PR #893, opened 2023-06-28
Allow signing substrate tx for moonbeam with evm account

chidg:
Fixes bug when attempting to sign Substrate transactions on Moonbeam on Polkadot JS Apps


alecdwm (approved):
LGTM

PR #899, opened 2023-06-29
fix: firefox i18next performance

### LINKED ISSUE #887: Popup loading performance Firefox
**Describe the bug**
In v 1.17.0, the extension popup is slow to load when opening in FF
This is much more pronounced when the dashboard is not also open

**To Reproduce**
Open the popup in FF, when the dashboard is not open

**Expected behaviour**
Popup loading should be equally fast when dashboard open or not open

0xKheops:
- Fixes #887 

Reduces the amount of translation files that have to be asynchronously loaded, to prevent a big performance hit on Firefox.

Namespaces are now organised by page area : 
- common - contains everything that needs to be displayed on portfolio and basic screens
- admin - all dashboard texts that aren't displayed in portfolio (settings, add accounts, about, etc.)
- send-funds - labels for send-funds, which has a dedicated popup
- onboard - labels displayed only in onboarding.html
- requests - labels used in signing requests, this is big and will grow

From now on we will need to pay attention when reviewing PRs that popup and dashboard home pages don't load any other namespace than "common". it's displayed in the console in dev mode, easy to check.


alecdwm (approved):
<div align="center">



</div>

PR #912, opened 2023-07-06
fix: ct-1036

0xKheops:


apps/extension/src/ui/domains/Account/AccountExportPrivateKeyModal.tsx
```diff
@@ -46,19 +45,10 @@ const ExportPrivateKeyResult = ({ onClose }: { onClose?: () => void }) => {
...
-  } = useQuery({
-    queryKey: ["accountExportPrivateKey", !!password],
-    queryFn: () => (password ? exportAccount(password) : null),
-    refetchInterval: false,
-    refetchOnWindowFocus: false,
-    refetchOnReconnect: false,
-    refetchOnMount: true,
-  })
+  // don't use react-query here as we don't want this to be cached
+  const [privateKey, setPrivateKey] = useState<string>()
```
chidg: It would be nice to have something like 
```
import { useState, useEffect } from 'react';

const useSensitiveState = (initialValue) => {
  const [value, setValue] = useState(initialValue);

  useEffect(() => {
    return () => {
      // Clear the sensitive state value on unmount
      setValue(null);
    };
  }, []);

  return [value, setValue];
};

export default useSensitiveState;
```

Might make it easier to remember to do this in future if we have a hook already

chidg (approved):
LGTM

PR #919, opened 2023-07-10
fix: autologin

0xKheops:
fixes devmode auto login that was broken in #912


alecdwm (approved):
LGTM!

Tested:
- no PASSWORD env var (doesn't autologin)
- incorrect PASSWORD env var (tries to autologin, fails, allows me to manually enter correct password)
- correct PASSWORD env var (does the autologin, doesn't seem to break anything)

PR #922, opened 2023-07-10
fix: allow extension builds with no SimpleLocalize api key

alecdwm:
Previously:
- Dev builds would fail (`process` is not defined errors in frontend and background script)
- Builds with no api key would fail (unable to fetch translations)
- Production builds with an api key would fail (`process` is not defined errors in frontend & background script)

I suspect the `process` is not defined errors might have been a race condition in the `SimpleLocalizeDownloadPlugin`, and is the reason I moved the download logic from the sync `afterPlugins` hook to the async `beforeRun` hook

After this PR:
- Dev builds / prod builds with no api key use english fallbacks
- Prod builds with api key have no `process` is not defined errors

### DIFF
--- a/apps/extension/i18next-parser.config.cjs
+++ b/apps/extension/i18next-parser.config.cjs
@@ -2,7 +2,10 @@
 
 // The key is the `locale` as passed to i18next.
 // The value is the `human-readable name` as passed to the language settings UI in the wallet.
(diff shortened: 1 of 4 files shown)


alecdwm:
Found an issue, fixing

chidg (dismissed):
LGTM

chidg (approved):
LGTM

PR #926, opened 2023-07-13
fix: sign substrate stuck analysing

### LINKED ISSUE #923: Talisman stuck on "Analysing Transaction" when signing TX from Polkadot dev node
When I attempt to sign a transaction using Talisman via a local Polkadot `--dev` node, it gets stuck on an "Analysing transaction, this shouldn't take long" screen and does not progress.

I'd like to try using Talisman against a local dev node in order to check that I can manually construct and sign extrinsics using it (specifically to work out how to construct and send the appropriate payload to the extension, ideally without using Polkadot.js).

**To Reproduce**

1. Install Polkadot, eg:
   ```
   git clone https://github.com/paritytech/polkadot.git
   cd polkadot
   cargo install --path .
   ```
    (Expects Rust/cargo to be installed)
1. Start a local polkadot dev node: `polkadot --dev`
1. With the Talisman extension enabled, open https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A9944#/extrinsics and attempt to submit some transaction using a Talisman account (I loaded the dev `Alice` account into Talisman to do this).

**Expected behavior**

The transaction should be signed by Talisman.

**Screenshots**

Instead, I see this indefinitely:

Screenshot 2023-07-10 at 14 53 05

**Version Information**

 - MacOS 11.6.8
 - Firefox 114.0.2
 - Talisman Extension 1.17.4
 - Polkadot 0.9.43-d5616884690 (the latest as of this issue; I also tried an older version to no avail)

0xKheops:
- Fixes #923 

We had an infinite loop situation because of nested components using the same useQuery hook.
Result is cached and shared properly, but errors weren't, prior to this PR.

### DIFF
--- a/apps/extension/src/ui/hooks/useExtrinsic.ts
+++ b/apps/extension/src/ui/hooks/useExtrinsic.ts
@@ -110,6 +110,7 @@ export const useExtrinsic = (payload?: SignerPayloadJSON | SignerPayloadRaw) => 
     refetchInterval: false,
     refetchOnWindowFocus: false,
     refetchOnReconnect: true,
+    retryOnMount: false,
     retry: 2,
   })
 }


chidg (approved):
LGTM

PR #927, opened 2023-07-14
Dependency upgrade 1.17

chidg:


apps/extension/src/core/util/isStringHexString.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @name isStringHexString
+ * @description Checks if a string is a hex string. Required to account for type differences between different polkadot libraries
+ * @param {string} str - string to check
+ * @returns {HexString} - hex string
+ * @example
+ * isStringHexString("0x1234") // "0x1234"
+ * isStringHexString("1234") // Error: Expected a hex string
+ * isStringHexString(1234) // Error: Expected a string
+ **/
+export const isStringHexString = (str: string): HexString => {
```
0xKheops: proposing to rename to `validateHexString` and change return type to `str is HexString` which would in practice return a boolean, and remove the need to get the result in a new variable when calling this method

0xKheops (changes requested):
It's working great, just got a method rename request :)

0xKheops (approved):
LGTM

PR #929, opened 2023-07-14
feat: account colors on popup home page

0xKheops:
- optional `colors property` on MysticalAccountV3's config, to force colors
- useFirstAccountColor hook that returns colors of first account (since now we can sort accounts, user can control which one it is)
- leverage useFirstAccountColor hook on both Login & TotalPortfolio

### DIFF
--- a/.changeset/large-laws-relax.md
+++ b/.changeset/large-laws-relax.md
@@ -1,0 +1,5 @@
+---
+"talisman-ui": minor
+---
+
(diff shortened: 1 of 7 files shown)


alecdwm:
I rebased this onto `dev`, since manage-accounts is now squash merged into it

alecdwm (approved):
LGTM!!!

PR #933, opened 2023-07-17
feat: protect window.ethereum from Phantom

### LINKED ISSUE #819: Phantom messes up with window.ethereum
Users are unhappy with "Talisman overrides metamask" but Talisman never does. 

Here is what happens : 
- Talisman injects in window.ethereum because it's empty (otherwise, it wouldn't)
- Phantom snapshots the value of window.ethereum (talisman)
- Phantom injects in window.ethereum and locks it up
- Metamask may try to override window.ethereum here, without success
- When Phantom fallbacks to metamask (after prompt or just because of user setting) it actually proxies requests to the snapshotted object, which is Talisman
- User blames Talisman

Note: this doesn't always happens. It depends on the order in which extensions inject, which seem to be the order in which they were installed/activated

### LINKED ISSUE #915: MM Ethereum window override issues
https://www.notion.so/talismanwallet/Fix-MM-Talisman-conflicts-becce52afa444410a156a466b7283cb8

### LINKED ISSUE #931: Fix Phantom/MM Conflicts

0xKheops:
- Fixes #819 
- Fixes #915 
- Fixes #931

### DIFF
--- a/apps/extension/src/core/injectEth/injectEthereum.ts
+++ b/apps/extension/src/core/injectEth/injectEthereum.ts
@@ -34,7 +34,7 @@ export const injectEthereum = (sendRequest: SendRequest) => {
     log.debug("Injecting talismanEth in window.ethereum")
 
     try {
-      // Protect window.ethereum - workaround Phantom abuse bug
+      // Protect window.ethereum property to workaround Phantom abuse bug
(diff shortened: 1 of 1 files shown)


chidg (approved):
LGTM

PR #936, opened 2023-07-18
fix: ui fixes

0xKheops:
Bundle of small fixes, following tailwind rampage : 
- Popup's zoom reset
- overflow issue in send funds confirm page
- button in button DOM error on total porfolio button => fixed using 2 containers on top of each other
- target of "Portfolio" button in sidebar
- NavLink invalid props
- classNames usage (overrides should be last)


alecdwm (approved):
LGTM!

PR #940, opened 2023-07-19
feat: sync watched accounts between extension and portal

tien:


apps/extension/src/core/constants.ts
```diff
@@ -31,7 +31,7 @@ export const API_KEY_ONFINALITY = process.env.API_KEY_ONFINALITY
 
 export const IPFS_GATEWAY = "https://talisman.mypinata.cloud/ipfs/"
 
-export const TALISMAN_WEB_APP_DOMAIN = "app.talisman.xyz"
+export const TALISMAN_WEB_APP_DOMAIN = DEBUG ? "localhost" : "app.talisman.xyz"
```
chidg: Let's have this provided in an env variable, with `app.talisman.xyz` as the default

apps/extension/src/ui/hooks/useAccounts.tsx
```diff
@@ -10,7 +10,7 @@ const INITIAL_VALUE: AccountJsonAny[] = []
 const subscribe = (subject: BehaviorSubject<AccountJsonAny[]>) =>
   api.accountsSubscribe((v) => subject.next(v))
 
-export type UseAccountsFilter = "all" | "watched" | "owned" | "portfolio"
+export type UseAccountsFilter = "all" | "watched" | "owned" | "portfolio" | "dapp" | "portal"
```
chidg: I don't understand the need for the two new filter types, when `owned` or `all` would have done the job?

chidg (changes requested):
Pretty good! Couple of small changes suggested.

apps/extension/src/core/constants.ts
tien: Make sense, would the name `PORTAL_DOMAIN` be sufficient of do we want the full `TALISMAN_WEB_APP_DOMAIN`?

apps/extension/src/core/handlers/Extension.ts
```diff
@@ -86,11 +86,7 @@ export default class Extension extends ExtensionHandler {
             .filter(([, site]) => site.connectAllSubstrate)
             .forEach(async ([url, autoAddSite]) => {
               const newAddresses = Object.values(addresses)
-                .filter(
-                  ({ json: { address, meta } }) =>
-                    meta.origin !== AccountTypes.WATCHED &&
-                    !autoAddSite.addresses?.includes(address)
-                )
+                .filter(({ json: { address } }) => !autoAddSite.addresses?.includes(address))
```
chidg: Because we might have `connectAllSubstrate` on other sites in future, we should still do the check for `meta.origin !== AccountType.Watched` here if the site is not the Portal

apps/extension/src/ui/hooks/useAccounts.tsx
tien: This is so:
- Developers doesn't have to keep a mental model of what accounts are available to DApp (it's in the function signature and definition). So less error-prone, easier understanding and debugging.
- Ability to add more subdivisions later on

apps/extension/src/core/handlers/Extension.ts
tien: Let me know if https://github.com/TalismanSociety/talisman/pull/940/commits/c80f187e4f1a355688f990ee9309e7998ba1452a works

apps/extension/src/core/constants.ts
alecdwm: We fixed this in another PR - maybe revert on this one so it doesn't conflict?

https://github.com/TalismanSociety/talisman/pull/995/files?w=1#diff-033eebf693f40c470c273d6f1866dc2d72d3e28f6d0dfb7901454d218fdb1352L42-R46

Or not - I can just rebase if this gets merged first

chidg (approved):
LGTM

PR #942, opened 2023-07-19
Move feature flags to github hosted TOML config

chidg:


apps/extension/src/core/handlers/Extension.ts
```diff
@@ -97,12 +101,43 @@ export default class Extension extends ExtensionHandler {
       DEBUG || TEST ? 0 : 2000
     )
 
+    // setup polling for config from github
+    let CONFIG_UPDATE_INTERVAL = 1000 * 60 * 5 // 5 minutes
+    const MAX_CONFIG_UPDATE_INTERVAL = 1000 * 60 * 60 // 1 hour
+    setInterval(() => {
+      try {
+        this.getRemoteConfig()
```
0xKheops: async function here, errors woudln't be caught.
might want to use prom.catch() pattern

apps/extension/src/core/handlers/Extension.ts
```diff
@@ -97,12 +101,43 @@ export default class Extension extends ExtensionHandler {
...
+      } catch (e) {
+        // exponential backoff for rate limits
+        if (
+          (e as Error).message === CONFIG_RATE_LIMIT_ERROR &&
+          CONFIG_UPDATE_INTERVAL < MAX_CONFIG_UPDATE_INTERVAL
+        )
+          CONFIG_UPDATE_INTERVAL = CONFIG_UPDATE_INTERVAL / 2
```
0xKheops: if exponential fallback i guess timeout must be increased

```suggestion
          CONFIG_UPDATE_INTERVAL = CONFIG_UPDATE_INTERVAL * 2
```

apps/extension/src/core/handlers/Extension.ts
```diff
@@ -97,12 +101,43 @@ export default class Extension extends ExtensionHandler {
...
     this.initDb()
     this.initWalletFunding()
     this.checkSpiritKeyOwnership()
     this.cleanup()
   }
 
+  private async getRemoteConfig() {
+    const config = await getConfig()
+
+    featuresStore.set({
```
0xKheops:
```suggestion
    await featuresStore.set({
```

0xKheops (changes requested):
found a problem on error & exponential fallback logic

also, i thought we were using onfinality api key for substrate still, but our logic prevented that as it was using the flag suffixed with _EVM.
I'm wondering if we should add it back by setting apikey on chaindataProvider object

apps/extension/src/core/handlers/Extension.ts
0xKheops: if method doesn't return anything, might want to rename it to `updateRemoteConfig`

chidg: renamed to `fetchRemoteConfig`

0xKheops (approved):
LGTM

PR #953, opened 2023-07-21
feat: inject private provider on staging portal

0xKheops:


apps/extension/src/core/page.ts
```diff
@@ -48,16 +48,12 @@ function inject() {
...
   injectEthereum(messageService.sendMessage)
 
-  // Custom `injectSub` endpoint used by app.talisman.xyz
-  // to access the ChainConnector instance of the wallet
-  const injectSubstrateHostnames = [
-    // always inject on https://app.talisman.xyz
-    TALISMAN_WEB_APP_DOMAIN,
+  // Inject restricted substrate provider for Talisman dapps
+  const regAllowPrivateProvider = DEBUG
+    ? /(^app.talisman.xyz$)|(.talisman.pages.dev$)|(^localhost$)/i
```
alecdwm: Afaik these `.` symbols are interpreted as `any character (except for line terminators)`

Which means the following hosts match the regex:
- `appatalisman.xyz`
- `appbtalisman.xyz`
- `app-talisman.xyz`
- etc

I think we should instead use something like this:
`(^app\.talisman\.xyz$)|(.+\.talisman\.pages\.dev$)|(^localhost$)`

Or alternatively, I'd vote we ditch regex and just do:
```ts
const testHostname = (hostname: string) =>
  (DEBUG && hostname === 'localhost') ||
  hostname === TALISMAN_WEB_APP_DOMAIN ||
  hostname.endsWith('.talisman.pages.dev')
```

Not as pretty as regex, but there's no hidden edge cases!

0xKheops: good catch! thanks

alecdwm (approved):
LGTM!

PR #961, opened 2023-07-24
fix: instruct vscode to ignore PR for dev branch

0xKheops:
Leverage a vscode setting that "Prevents branches that are associated with a pull request from being automatically detected. This will prevent review mode from being entered on these branches."

Since dev was branched off main, vscode keeps thinking that we intend to make a PR from it. This setting fixes that.


alecdwm (approved):
Nice!

PR #962, opened 2023-07-24
Better error handling in getConfig

chidg:
Handle errors in fetching to prevent Sentry error TALISMAN-EXTENSION-T1Z

### DIFF
--- a/apps/extension/src/core/util/getConfig.ts
+++ b/apps/extension/src/core/util/getConfig.ts
@@ -1,5 +1,6 @@
 import { TALISMAN_CONFIG_URL } from "@core/constants"
 import { FeatureVariants } from "@core/domains/app/types"
+import { log } from "@core/log"
 import toml from "toml"
 
 type Config = {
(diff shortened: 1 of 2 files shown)


apps/extension/src/core/util/getConfig.ts
```diff
@@ -8,8 +9,15 @@ type Config = {
 
 export const CONFIG_RATE_LIMIT_ERROR = "Rate Limit Exceeded"
 
-export const getConfig = async (): Promise<Config> => {
-  const response = await fetch(TALISMAN_CONFIG_URL)
+export const getConfig = async (): Promise<Config | null> => {
+  try {
+    // eslint-disable-next-line no-var
+    var response = await fetch(TALISMAN_CONFIG_URL)
```
alecdwm, line 15: Something to check - I don't think that `fetch` raises an error when the response is not `HTTP 200`.

So we might want another `if` which throws an error if [response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) is false.

Otherwise we'll see toml parser errors when the network / github is down.

chidg: Correct, i'll add something for that

0xKheops (approved):
LGTM

PR #964, opened 2023-07-24
chore: add license to monorepo root

alecdwm:
I noticed we have the license on every package in this repo, but not on the repo itself.

I was hoping to change this so that I can use this badge:

[license](https://github.com/TalismanSociety/talisman/blob/dev/LICENSE)


0xKheops (approved):
LGTM !

PR #965, opened 2023-07-24
fix: delete LICENSE for package which doesn't exist

alecdwm:
We have a LICENSE file at `packages/talisman-utils/LICENSE` but `talisman-utils` doesn't exist.

This PR deletes it.

It also fixes the CI, which before this PR would break if we try and remove any packages from the repo.


0xKheops (approved):
LGTM

PR #973, opened 2023-07-28
fix: persist collapsed folders in dashboard account selector

alecdwm:


apps/extension/src/core/domains/app/store.settings.ts
```diff
@@ -9,6 +9,7 @@ export interface SettingsStoreData {
   hideBalances: boolean
   allowNotifications: boolean
   selectedAccount?: string // undefined = show all accounts
+  collapsedFolders?: string[] // persists the collapsed folders in the dashboard account picker
```
chidg, line 12: Would it be possible to store this in the accounts catalog store rather than the settings?

chidg (approved):
LGTM

PR #975, opened 2023-07-29
fix: popup size issues

### LINKED ISSUE #963: Padding issue on Firefox
Incorrect right-padding on firefox, non reproductible on chrome.
Identified on conviction voting sign view, but might be present on other screens

image

0xKheops:
- Fixes #963 

#963 was due to the fact that there may be a delta between window.innerWidth and window.outerWidth on some OS/browsers. This is also true for height.

Details :
- This PR will resize the popup when it loads to take the size delta into account, if and only if necessary.
- Since this delta is always the same for a given OS/browser, it will be stored in app state so next popups can be open directly with the perfect size
- Also found a way to reset zoom only for the popup, without touching dashboard's zoom level. this works only on chromium browsers though.

Note : zoom still has to be reset, otherwise the size delta cannot be adjusted accurately. tried with a lot of formulas, and sadly i didn't find one that works.

Tested on : 
- [x] Windows chrome, brave and firefox - all with & without OS custom scale
- [x] Linux chrome, brave and firefox
- [ ] MacOS chrome, brave, direfox and Arc

Could anyone test & if necessary adjust this for Arc please ?

### DIFF
--- a/.changeset/quiet-ears-remain.md
+++ b/.changeset/quiet-ears-remain.md
@@ -1,0 +1,5 @@
+---
+"talisman-ui": patch
+---
+
+popup borders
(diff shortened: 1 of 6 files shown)


chidg (approved):
LGTM

PR #976, opened 2023-07-31
feat: dcent

### LINKED ISSUE #958: D'cent Integration

0xKheops:
- Closes #958 

Misc : 
- Can now sign transactions on legacy evm gas networks (ex: send BNB) with ledger
- Refactored Ledger signing components so they share same props as D'Cent signing components (for both evm & substrate)

Note : most commits are marked as unverified because pushed from my Windows VM. Didn't take time to configure GitHub SSH.


apps/extension/src/core/domains/balances/types.ts
```diff
@@ -21,9 +21,14 @@ export type AddressesByEvmNetwork = {
...
   addresses: string[]
 }
+export type AddressesByTokens = {
+  tokenIds: TokenId[]
+  addresses: string[]
+}
 export interface RequestBalancesByParamsSubscribe {
   addressesByChain: AddressesByChain
   addressesByEvmNetwork: AddressesByEvmNetwork
+  addressesByTokens: AddressesByTokens
```
chidg: Slightly confused by the shape of this type given the name - I notice that `AddressesByEvmNetwork` is a different shape to `AddressesByChain` and you're following that with `AddressesByTokens`. IMO a variable with name `xByY` should have a shape like `Record<y, x[]>`. Not a blocker but would be nice to make consistent in future.

apps/extension/src/ui/domains/Sign/SignDcentEthereum.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+const signWithDcent = async (
+  method: EthSignMessageMethod | "eth_sendTransaction",
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  payload: any,
+  accountPath: string
+): Promise<HexString> => {
+  switch (method) {
+    // case "eth_signTypedData":
+    // case "eth_signTypedData_v1":
```
chidg: These commented methods are not compatible with Dcent? If so maybe you should add an additional comment to explain

chidg (dismissed):
Comments not blocking... LGTM

chidg (approved):
LGTM

PR #977, opened 2023-07-31
fix: dev mode watch ignore paths

0xKheops:


chidg:
Much faster reloads, thanks for noticing this!

chidg (approved):
LGTM

PR #978, opened 2023-07-31
feat: add polkadot faucet link to testnets toggle

alecdwm:


0xKheops:
might want to provide a link for westend faucet too, at first i didn't see the dropdown in the top right. Or maybe it's just the link name that tricked me into thinking the link is rococo specific.

So either 2 links, or rename the single link from "Rococo Faucet" to "Faucets".
westend direct link : https://paritytech.github.io/polkadot-testnet-faucet/westend

0xKheops (approved):
LGTM, tho I think we should provide link(s) to EVM faucet(s) too at some point

PR #980, opened 2023-08-01
Make initial popup height delta 30 when IS_FIREFOX true

chidg:
Fixes some sizing quirks:
- Some browser/OS combinations include browser title bar as part of 'window.outerHeight'
- Firefox does not allow you to resize a window unless created by the `window.open` method, which it doesn't allow in extensions. This PR users the `Browser.windows.update` api to get around this .


0xKheops (approved):
LGTM

PR #992, opened 2023-08-04
fix: don't resize popup in normal window

0xKheops:
- don't resize the popup if open in a normal window (useful for devs)
- if opening a popup breaks, most likely because of invalid size deltas, retry with default size

### DIFF
--- a/apps/extension/src/index.popup.tsx
+++ b/apps/extension/src/index.popup.tsx
@@ -13,7 +13,13 @@ const adjustPopupSize = async () => {
   if (window.location.search === "?embedded") return
 
   try {
-    const currentZoom = await Browser.tabs.getZoom()
+    const [currentWindow, currentZoom] = await Promise.all([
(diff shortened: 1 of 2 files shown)


apps/extension/src/index.popup.tsx
```diff
@@ -13,7 +13,13 @@ const adjustPopupSize = async () => {
...
 
   try {
-    const currentZoom = await Browser.tabs.getZoom()
+    const [currentWindow, currentZoom] = await Promise.all([
+      Browser.windows.getCurrent(),
+      Browser.tabs.getZoom(),
+    ])
+
+    // exit if popup is opened in a normal window (common for devs)
+    if (currentWindow.type !== "popup") return
```
alecdwm, line 22: 🤩

alecdwm (approved):
<div align="center">

awesome

</div>

PR #995, opened 2023-08-07
fix: some wallet QOL fixes

alecdwm:
Some commits which have been sitting around in my local dev env:

- https://github.com/TalismanSociety/talisman/commit/505e61b20eba2c6166494c4d20d77aa3c9a5cb35 adds 127.0.0.1 to the list of hosts we should inject `talismanSub` into for webapp development
- https://github.com/TalismanSociety/talisman/commit/ffd67e775502480d0568a208dafbb2e02bd87516 goes through and removes all hardcoded webapp URLs, moving them into the constants.ts file (which will soon be mostly offloaded to [talisman-config](https://github.com/talismansociety/talisman-config))
- https://github.com/TalismanSociety/talisman/commit/b3d8c8bb073407619b994d4769349871a0bdd06e stops github from rendering the full diff of the yarn sourcecode when we upgrade yarn / any yarn plugins
- https://github.com/TalismanSociety/talisman/commit/28d059336dc6b07baedb69460d3b528c54ce0cca fixes some performance bugs I noticed in the login flow, details below:

**In the backend**
We were hashing the password twice for every login by:
1. Calling `stores.password.transformPassword` - which we only use the result of when migrating users to the new login structure
2. Calling `stores.password.authenticate`, which internally calls `stores.password.transformPassword` as well

I ran a performance test before changing anything, and for me the double-call to the hash function caused a 2x increase in waiting before the authorisation completed (which makes sense). From approx 1000ms to 2000ms on my machine.

**In the frontend**
Only applies to the dev autologin, but annoyingly it would call the authorise backend function three times on load!
This would cause the backend to run the same hash three times in parallel - slowing the completion of all three of the authorisation calls.
On my machine this makes a 3x difference to the time spent waiting - from approx 1000ms for a single hash call, to approx 3000ms.

With the two fixes combined, this PR brings the login time from about 6000ms to 1000ms.

### DIFF
--- a/apps/extension/src/core/domains/app/handler.ts
+++ b/apps/extension/src/core/domains/app/handler.ts
@@ -131,7 +131,7 @@ export default class AppHandler extends ExtensionHandler {
         pair.lock()
 
         // we can now set up the auth secret
-        await this.stores.password.setPassword(transformedPassword)
+        this.stores.password.setPassword(transformedPassword)
         await this.stores.password.setupAuthSecret(transformedPassword)
       } else {
         await this.stores.password.authenticate(pass)
--- a/apps/extension/src/core/domains/app/handler.ts
+++ b/apps/extension/src/core/domains/app/handler.ts
@@ -111,14 +111,13 @@ export default class AppHandler extends ExtensionHandler {
   }
 
   private async authenticate({ pass }: RequestLogin): Promise<boolean> {
-    await new Promise((resolve) =>
-      setTimeout(resolve, process.env.NODE_ENV === "production" ? 1000 : 0)
-    )
+    if (!(DEBUG || TEST)) await sleep(1000)
 
     try {
-      const transformedPassword = await this.stores.password.transformPassword(pass)
       const { secret, check } = await this.stores.password.get()
       if (!secret || !check) {
+        const transformedPassword = await this.stores.password.transformPassword(pass)
+
         // attempt to log in via the legacy method
         const primaryAccount = getPrimaryAccount(true)
         assert(primaryAccount, "No primary account, unable to authorise")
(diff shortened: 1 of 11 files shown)


0xKheops (approved):
Nice, LGTM !

apps/extension/src/core/domains/app/handler.ts
```diff
@@ -132,7 +131,7 @@ export default class AppHandler extends ExtensionHandler {
         pair.lock()
 
         // we can now set up the auth secret
-        await this.stores.password.setPlaintextPassword(pass)
+        this.stores.password.setPassword(transformedPassword)
         await this.stores.password.setupAuthSecret(transformedPassword)
       } else {
         await this.stores.password.authenticate(pass)
```
alecdwm, line 137: For context, this is the implementation of `this.stores.password.authenticate`:

```ts
  async authenticate(password: string) {
    if (this.isLoggedIn.value === TRUE) return
    const pw = await this.transformPassword(password)
    const { secret, check } = await this.get()
    assert(secret && check, "Unable to authenticate")
    const result = (await decrypt(pw, check)) as { secret: string }
    assert(result.secret && result.secret === secret, "Incorrect Password")
    this.setPassword(pw)
  }
  ```

File: apps/extension/src/core/domains/app/handler.ts [hunk 2]
```diff
@@ -132,7 +131,7 @@ export default class AppHandler extends ExtensionHandler {
         pair.lock()
 
         // we can now set up the auth secret
-        await this.stores.password.setPlaintextPassword(pass)
+        this.stores.password.setPassword(transformedPassword)
```
alecdwm, line 134: For context, this is the implementation of `this.stores.password.setPlaintextPassword`:

```ts
  public async setPlaintextPassword(plaintextPw: string) {
    const pw = await this.transformPassword(plaintextPw)
    this.setPassword(pw)
  }
  ```
  
  We've already derived a transformedPassword, so why don't we just call `setPassword` and avoid running another 1s hash?

chidg (approved):
LGTM!

PR #998, opened 2023-08-08
fix: account icons

0xKheops:
- Fixes incorrect account icon size in import from seed screen and address book screen
- Adds missing network badges to the dashboard account dropdown and to the manage accounts screen


alecdwm (approved):
Nice!

PR #1002, opened 2023-08-11
fix: use correct favicons everywhere

alecdwm:
Fixes the wrong favicon showing in the reminder to pin the extension, and also next to the tab title.


apps/extension/src/template.dashboard.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="./favicon.svg" />
+    <link rel="icon" href="./favicon.svg?v=<%= __webpack_hash__  %>" />
```
alecdwm, line 5: Browsers aggressively cache favicons, this hash is needed in order to get it to update between dev and canary builds (so when the extension id doesn't change).

apps/extension/src/@types/react-app-env.d.ts
```diff
@@ -12,7 +12,7 @@ declare namespace NodeJS {
     readonly SENTRY_AUTH_TOKEN?: string
     readonly SIMPLE_LOCALIZE_API_KEY?: string
     readonly SIMPLE_LOCALIZE_PROJECT_TOKEN?: string
-    readonly BUILD?: "production" | "canary" | "ci"
+    readonly BUILD?: "production" | "canary" | "ci" | "qa" | "dev"
```
alecdwm, line 15: Doesn't change anything, the type was just missing two possible values

apps/extension/webpack/webpack.prod.js
```diff
@@ -35,10 +39,6 @@ const config = (env) => {
             from: "manifest.json",
             to: distDir,
             context: "public",
-            // overwrite non-transformed manifest.json
```
alecdwm, line 38: Not needed when we don't copy the manifest.json twice

0xKheops (approved):
Cool one, thanks !
LGTM !

PR #1006, opened 2023-08-14
feat: onchain id

### LINKED ISSUE #982: Show address for address book contact in send funds dialog
Currently when selecting an address book contact from the recipients list in the send funds UI, the contact name is displayed but not the address. This can lead to a situation where if a user has multiple contacts with the same or similar names, they could select the wrong one.

We can make this more clear by showing the address in two places:
1. In the recipients list, on a second line underneath the name of the contact
Screenshot 2023-08-02 at 10 06 19 am

2. In the Amount entry screen, as a tooltip on the recipient component
Screenshot 2023-08-02 at 10 06 28 am

To be completed in conjunction with #960

alecdwm:
Adds ENS lookup support to the wallet.

Includes functionality to query polkadot identities, but does not expose them in the UI.


lee-qa:
Tested and LGTM. Suggestion is to change the wallet address input placeholder text to "Address or ENS" or "Address or domain". 

Tested ENS on modules:
- Add Watched Account
- Add contact
- Receive Funds modal
- Send Funds modal

packages/on-chain-id/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    const provider = await this.chainConnectors.evm?.getProviderForEvmNetworkId(
+      this.networkIdEthereum
+    )
+    if (!provider) {
+      log.warn(`Could not find Ethereum provider in OnChainId::lookupEnsAddresses`)
+      return onChainIds
+    }
+
+    const { errors } = await PromisePool.withConcurrency(3)
```
0xKheops: Might not need to use PromisePool as our EVM provider is batching requests.

packages/on-chain-id/src/helpers.ts
```diff
@@ -0,0 +1 @@
+export const isPotentialEns = (name?: string) => typeof name === "string" && /^.+\..+$/.test(name)
```
0xKheops, line 1: shouldn't it check that the string ends with `.eth` ?

apps/extension/src/ui/domains/SendFunds/SendFundsAmountForm.tsx
```diff
@@ -86,14 +87,16 @@ const AddressPillButton: FC<AddressPillButtonProps> = ({ address, className, onC
...
+
   if (!address) return null
 
   return (
     <PillButton className={classNames("h-16 max-w-full !px-4", className)} onClick={onClick}>
       <div className="text-body flex h-16 max-w-full flex-nowrap items-center gap-4 overflow-x-hidden text-base">
         <AccountIcon className="!text-lg" address={address} genesisHash={genesisHash} />
         <div className="leading-base grow overflow-hidden text-ellipsis whitespace-nowrap">
-          {name ?? shortenAddress(address, 6, 6)}
+          {name ?? <Address address={formattedAddress} startCharCount={6} endCharCount={6} />}
```
0xKheops: Tooltip from [#982](https://github.com/TalismanSociety/talisman/issues/982) seems to be working only for ENS addresses

apps/extension/src/ui/domains/Account/AddressFieldEnsBadge.tsx
```diff
@@ -0,0 +1,43 @@
...
+    {isEnsLookup && ensLookup && (
+      <WithTooltip
+        tooltip={ensLookup}
+        // if we're not in a popup, no need to wrap
+        noWrap={!document.getElementById("main")}
+      >
+        <PillButton className="h-16 max-w-full !cursor-default !px-4">
+          <div className="text-body flex h-16 max-w-full flex-nowrap items-center gap-4 overflow-x-hidden text-base">
+            <AccountIcon className="text-lg" address={ensLookup} />
+            <div className="leading-base grow overflow-hidden text-ellipsis whitespace-nowrap">
```
0xKheops: handy shortcut :)

```suggestion
            <div className="leading-base grow truncate">
```

apps/extension/src/ui/hooks/useResolveEnsName.ts
```diff
@@ -0,0 +1,35 @@
+import { isEthereumAddress } from "@polkadot/util-crypto"
+import { isPotentialEns } from "@talismn/on-chain-id"
+import {
+  ensNamesIsFetching,
+  readEnsNameLookups,
+  useResolveAddressesForEnsNames,
+} from "@ui/atoms/ensNames"
+import { useRecoilValue } from "recoil"
+
+export const useResolveEnsName = (name?: string) => {
```
0xKheops, line 6: I may be missing something but this + ensNames.ts (mainly) feels overengineered and hard to read/maintain.
wdyt about using a useQuery call instead ?

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioAssets.tsx
```diff
@@ -114,8 +123,17 @@ const PageContent = ({
...
-            <div className="text-md overflow-hidden text-ellipsis whitespace-nowrap">
-              <Fiat amount={balances.sum.fiat("usd").total} isBalance />
+            <div
+              className={classNames(
+                "overflow-hidden text-ellipsis whitespace-nowrap",
+                account ? "text-body-secondary" : "text-md"
+              )}
+            >
+              {account ? (
+                <Address address={formattedAddress} />
```
0xKheops, line 118: thinking it would be more practical if the `<Address />` component was resolving addresses on it's own, with a flag to enable or disable this.
We got addresses all over the place, and will have many more in the future, so it would be nice not to have to resolve chainIds upfront each time we want to display an address, nor even to have to think about it.

0xKheops (commented):
few suggestions below that I would like to discuss :)

chidg:
@alecdwm could you please update on your progress against @0xKheops comments, and fix conflicts?

packages/on-chain-id/src/helpers.ts
alecdwm: [No ser](https://docs.ens.domains/dapp-developer-guide/resolving-names)

> a correct integration of ENS treats any dot-separated name as a potential ENS name and will attempt a look-up

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioAssets.tsx
alecdwm: I think I don't understand what you mean. It looks to me like the `<Address />` component *is* resolving addresses on its own?

apps/extension/src/ui/hooks/useResolveEnsName.ts
alecdwm: > feels overengineered and hard to read/maintain

guilty

> I may be missing something

I think I went with the atom (1) for [the localStorage persistence](https://github.com/TalismanSociety/talisman/pull/1006/files?w=1#diff-df5b6206e34200d1bf39b3f795f78d853307bb2c205ab9f54d59b7c8afc30901R127-R131) and (2) so that I could make a single request to the backend containing all of the names to query.

I found a useQuery solution for (1) and I think (2) doesn't matter anymore since we ditched the `PromisePool`.

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioAssets.tsx
0xKheops: right nevermind, thought the useFormattedAddress call that is often called upfront was doing some resolving

apps/extension/src/ui/atoms/onChainIds.ts
```diff
@@ -0,0 +1,127 @@
...
+    setMounted(true)
+    return () => setMounted(false)
+  }, [addresses, setMounted])
+}
+
+//
+// selectors
+//
+
+export const readOnChainIds = selectorFamily({
```
0xKheops: feeling the same with this one, wdyt about replacing it with a useQuery call ?

apps/extension/src/ui/hooks/useResolveEnsName.ts
```diff
@@ -0,0 +1,51 @@
...
+      // persist cache to local storage
+      persistEnsNamesCache()
+    },
+  })
+
+  return [address, { isLookup, ...rest }] as const
+}
+
+const ensNamesCache = new Map<string, string | null>(
+  JSON.parse(localStorage.getItem("TalismanEnsNamesCache") ?? "[]")
```
0xKheops: feeling this would fit better in a dexie table as it's "throwable data" and wouldn't block ui thread, though i understand that the simplicity of using localStorage is a benefit.
curious to know @chidg opinion on this ?

0xKheops (commented):
wondering about 2 minor things

apps/extension/src/ui/hooks/useResolveEnsName.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useQuery } from "@tanstack/react-query"
+import { api } from "@ui/api"
+
+export const useResolveEnsName = (name?: string) => {
+  // check if name is something we can look up
+  const lookupName =
+    // don't look up undefined
+    name !== undefined &&
+    // don't look up ethereum addresses
+    !isEthereumAddress(name) &&
```
chidg, line 12: Any chance we could use a more simple method like regex or string matching to identify an eth address here? `isEthereumAddress` from polkadot utils performs a cryptographic check on the value and is big and heavy. Ultimately I'd like to remove it from the frontend of the extension. We have `@talisman/util/isEthereumAddress`.

apps/extension/src/ui/hooks/useResolveEnsName.ts
chidg: Throwable in the sense that this could throw an error? Personally I am not a big fan of indexedDB and dexie in general, it has some specific features that are valuable for some use cases for us (subscriptions, big storage) but it generates a lot of errors that are hard to troubleshoot and handle. For small data objects which don't change very often like this, I think localstorage is a good solution. Is there another way we can get around the blocking issue @0xKheops ?

0xKheops: i meant throwable data as data that can be deleted and regenerated in case there is too low storage space.
it's a similar case as balances and other data that we put in indexed db currently, where wallet doesn't mind if browser purges the db.

The ui blocking is because local storage read/writes are synchronous, not a huge deal tho as cache is loaded only once each time the dashboard or popup is opened. Might be worth checking if there's a perf hit with 20+ accounts or a big address book but probably not. 
workaround would be dexie or extension storage which use asynchronous read/writes, or just no  caching at all on frontend appart from useQuery's : maybe in-memory cache on backend side would be sufficient

it was mainly food for thoughts, current implementation works.

apps/extension/src/ui/atoms/onChainIds.ts
alecdwm: I like how this one collects all the addresses to be fetched into one place, but I think in the future this should be done inside the `@talismn/on-chain-ids` lib instead of on the frontend

apps/extension/src/ui/hooks/useResolveEnsName.ts
alecdwm: I think this is more likely to become a problem if the cache grows forever, so I've made a fix for that!

chidg (approved):
LGTM!!

PR #1007, opened 2023-08-14
fix: EVM transaction gas improvements

### LINKED ISSUE #753: Estimated gas to low

### LINKED ISSUE #994: Validate balance is higher than max fee in EVM transaction
When sending funds or signing a transaction on EVM networks, we should raise a nice error message if the account balance is lower than the max fee

In send funds, while this may be validated before the review screen, if the gas price spikes in the time between showing the review screen and submitting the transaction, the request will fail with a generic error message. We should catch this and show a specific error message.

0xKheops:
- Fixes #753 
- Closes #994 

Fixes : 
- Adds a 2% safety margin on gas limit
- EVM transactions will be considered invalid if balance is insufficient to pay for gas
- In EVM transaction request popup, gas tooltip will now display balance instantly

Until now we were using the estimated gas as gas limit.
The gas cost of a contract call can change overtime if it's logic is based on any oracle or state that can change, so we need to add a safety margin to that estimate when setting the gas limit.

=> Pirates may now voyage without tweaking gas settings

pirate life

### DIFF
--- a/apps/extension/src/core/domains/ethereum/helpers.ts
+++ b/apps/extension/src/core/domains/ethereum/helpers.ts
@@ -128,16 +128,12 @@ export const getGasLimit = (
 ) => {
   // some dapps use legacy gas field instead of gasLimit
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const suggestedGasLimit = tx?.gasLimit ?? (tx as any)?.gas
-
-  const bnSuggestedGasLimit = suggestedGasLimit
-    ? BigNumber.from(suggestedGasLimit)
-    : BigNumber.from(0)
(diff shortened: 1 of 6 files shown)


chidg (approved):
LGTM

PR #1008, opened 2023-08-15
fix: bump min chrome version to 98

### LINKED ISSUE #1005: Upgrade chrome minimum version in manifest to 98
Some users have reported bugs regarding use of `structuredClone` in older versions of Chrome. `structuredClone` is only available from Chrome v98 https://caniuse.com/?search=structuredClone.

Update the extension manifest file to include Chrome 98 as the minimum version.

0xKheops:
- Closes #1005 

Minimum versions with structuredClone support : 
- Chrome 98
- Firefox 94

### DIFF
--- a/apps/extension/public/manifest.json
+++ b/apps/extension/public/manifest.json
@@ -39,5 +39,5 @@
       "strict_min_version": "95.0"
     }
   },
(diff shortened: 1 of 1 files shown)


chidg (approved):
LGTM

PR #1011, opened 2023-08-16
fix: evm failing providers recovery

0xKheops:
Currently if an EVM provider has no healthy rpc, null is set in the provider cache for that network.
The network then can't be used until browser is restarted.

This PR introduces a cleanup every 30 seconds for null values in the provider cache, allowing extension to retry to connect to failing networks after a short while.

### DIFF
--- a/.changeset/yellow-kangaroos-clean.md
+++ b/.changeset/yellow-kangaroos-clean.md
@@ -1,0 +1,5 @@
+---
+"@talismn/chain-connector-evm": patch
+---
+
+fix: allow failing providers to recover by removing them from cache every 30 seconds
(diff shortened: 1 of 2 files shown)


alecdwm:
What do you think about this, instead of the `setInterval` @0xKheops?

alecdwm:
Important detail, the race will always resolve to `null` when it's available - it won't randomly choose either `null` or `undefined`:

> [If the iterable contains one or more non-promise values and/or an already settled promise, then Promise.race() will settle to the first of these values found in the iterable.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race#description)

0xKheops:
> What do you think about this, instead of the `setInterval` @0xKheops?

yes

Let's go!

alecdwm (approved):
Let's gooooo

PR #1012, opened 2023-08-16
fix: custom token display network error

### LINKED ISSUE #857: [ISSUE] When adding custom tokens, error message is not descriptive
**Describe the bug**
Cannot add custom token to the wallet, most likely due to RPC limit

**To Reproduce**

1. Tried adding a custom token using settings, but it did not pick up any token, and said the contract was incorrect. 
2. Then I went on coingecko, and use the MM icon to add the token, and this worked OK (same contract address in extrinsic)

**Expected behaviour**
Kheops says this is likely an RPC rate limit issue, and so we should show an appriopriate error here instead so that they user does not get confused.

**Links**
The token (but it has same issue with other tokens too)
https://www.coingecko.com/en/coins/verse

**Screenshots**
If applicable, add screenshots/recordings to help explain your problem.
verse
vers2

**Environment (please complete the following information):**

- OS: iOS
- Browser Brave
- Version 1.6.4

0xKheops:
- Fixes #857 

If there is a connection issue, message will be "Failed to connect" instead of "Invalid address"

### DIFF
--- a/apps/extension/src/ui/apps/dashboard/routes/Tokens/AddCustomTokenPage.tsx
+++ b/apps/extension/src/ui/apps/dashboard/routes/Tokens/AddCustomTokenPage.tsx
@@ -115,6 +115,14 @@ export const AddCustomTokenPage = () => {
     [navigate, tokenInfo]
   )
 
+  const addressErrorMessage = useMemo(() => {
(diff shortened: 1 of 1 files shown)


chidg (approved):
LGTM

PR #1013, opened 2023-08-16
feat: copy ledger address skip chain selection

### LINKED ISSUE #780: When copying address for substrate ledger account, skip network selection step
Ledger accounts on substrate are only valid for one network. When there is only one network in the list of possible networks in the copy address modal, we should skip the network selection and go straight to displaying the address.

0xKheops:
- Fixes #780


alecdwm (approved):
Awesome!

PR #1014, opened 2023-08-16
fix: validation for addEthereumChain

0xKheops:
Removed the usage of `.url()` test from yup as it didn't validate a perfectly valid rpc url (hosted locally)
Also added the no script check to all strings as they are displayed in the view details drawer.


chidg (approved):
Seems like we could build a better, more permissive regex that would give the URL validation we need
but either way
LGTM!

PR #1016, opened 2023-08-16
Multiple seedstore migration

chidg:
For testing:
When upgrading from 1.16.4 or older:
- [ ] upgrade when no VC mnemonic set yet
  - [ ] expect that no VC mnemonic would be set after migration
 
When upgrading from 1.18.2:
- [ ] upgrade when using Talisman VC mnemonic
  - [ ] expect that the Talisman mnemonic ID would be stored as the VC mnemonic
- [ ] upgrade when using specific VC mnemonic
  - [ ] expect that the specific mnemonic ID would be stored as the VC mnemonic
- [ ] upgrade when set to not use a VC mnemonic at all
  - [ ] expect that `null` would be stored as the VC mnemonic


apps/extension/src/core/domains/mnemonics/store.ts
```diff
@@ -0,0 +1,130 @@
...
+  Generated = "generated",
+  Legacy = "legacy",
+  Vault = "vault",
+}
+
+export type SeedPhraseData = {
+  id: string
+  name: string
+  cipher?: string
+  source: SOURCES
```
0xKheops, line 21: Not sure about the Vault and Legacy values, as we wouldn't know if they have been generated or imported.
Do we need a source field at all, since we have the `confirmed` flag ?

apps/extension/src/core/domains/mnemonics/store.ts
```diff
@@ -0,0 +1,130 @@
...
+  } catch (e) {
+    log.error(e)
+    return Err(MnemonicErrors.UnableToDecrypt)
+  }
+  return Ok(mnemonic)
+}
+
+type DecryptedMnemonic = LegacySeedObj | string
+
+export class SeedPhraseStore extends StorageProvider<SeedPhraseStoreData> {
```
0xKheops, line 58: wdyt about replacing all "Seed" occurences by "Mnemonic" in class, variable, function and properties names ? (there are a bunch)

apps/extension/src/core/domains/accounts/handler.ts
```diff
@@ -488,7 +491,7 @@ export default class AccountsHandler extends ExtensionHandler {
       case "pri(accounts.validateMnemonic)":
         return this.accountValidateMnemonic(request as string)
       case "pri(accounts.setVerifierCertMnemonic)":
```
0xKheops, line 493: wdyt about moving this to the mnemonics handler ?

apps/extension/src/ui/hooks/useSeedPhrases.ts
```diff
@@ -0,0 +1,33 @@
...
+export const seedPhraseState = atom<SeedPhrases>({
+  key: "seedPhraseState",
+  default: {},
+  effects: [
+    ({ setSelf }) => {
+      const sub = seedPhraseStore.observable.subscribe((data) => {
+        const result = Object.fromEntries(
+          Object.entries(data).map(([id, seed]) => {
+            const { name, confirmed, source } = seed
+            return [id, { name, confirmed, source }]
```
0xKheops, line 21: thinking a simple array would be more convenient.
not having the id in values makes it hard to get, for example, the id of the first mnemonic that is not confirmed

apps/extension/src/ui/apps/dashboard/layout/DashboardNotifications/BackupNotification.tsx
```diff
@@ -49,13 +50,17 @@ export const BackupWarningModal = ({ isOpen, snooze, openBackupModal }: BackupWa
 export const BackupNotification = () => {
   const { t } = useTranslation()
   const { isOpen, open, close } = useOpenClose()
-  const { isNotConfirmed, isSnoozed, showBackupWarning, snoozeBackupReminder } = useMnemonicBackup()
+  const { allBackedUp, isSnoozed, showBackupWarning, snoozeBackupReminder } = useMnemonicBackup()
+
+  // todo need a way to choose mnemonic Id, for now just use the first
+  const mnemonics = useSeedPhrases()
+  const firstMnemonicId = Object.keys(mnemonics)[0]
```
0xKheops, line 57: I guess we shouldn't backup from here anymore, but instead redirect to upcoming manage seeds screen

0xKheops (commented):
seems good, just minor comments

apps/extension/src/ui/apps/dashboard/layout/DashboardNotifications/BackupNotification.tsx
chidg: Not a problem, these changes were hacky and only intended as first drafts - so feel free to take it and run with it

lee-qa:
Tested these scenarios with test results:

When upgrading from 1.16.4 or older:
- upgrade when no VC mnemonic set yet - OK

When upgrading from 1.18.2:
- upgrade when using Talisman VC mnemonic - OK
- upgrade when using specific VC mnemonic - OK
- upgrade when set to not use a VC mnemonic at all - "vaultVerifierCertificateMnemonicId" property is not shown and unable to verify null value

apps/extension/src/core/domains/mnemonics/store.ts
chidg: Removed 'vault' in https://github.com/TalismanSociety/talisman/pull/1050. I think we should keep the `source` field, since it might be useful to show whether a mnemonic is imported or created at some point.

apps/extension/src/core/domains/mnemonics/store.ts
chidg: WIP in https://github.com/TalismanSociety/talisman/pull/1050

apps/extension/src/ui/hooks/useSeedPhrases.ts
chidg: I see you changed it, I would have preferred to keep the object structure but just add the id as one of the properties, to prevent unecessary loops.

chidg:
Closing - this branch is part of #1031, with further updates completed in #1050

PR #1020, opened 2023-08-16
feat: multi-currency support

### LINKED ISSUE #986: Multi-Currency

tien:


apps/extension/src/ui/apps/dashboard/routes/Settings/CurrencySettingsPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+                {currencyConfig[currency]?.unicodeCharacter ?? ""}{" "}
+                {currencyConfig[currency]?.name ?? currency}
+              </div>
+            </div>
+            {selectableCurrencies.includes(currency) ? (
+              <StarIcon className="stroke-primary fill-primary text-lg" />
+            ) : (
+              <StarIcon className="text-lg" />
+            )}
+          </button>
```
0xKheops: would be more consistent with other settings screens
```suggestion
          <button
            type="button"
            key={currency}
            className="bg-grey-850 enabled:hover:bg-grey-800 text-body-disabled enabled:hover:text-body-secondary flex h-28 w-full cursor-pointer items-center gap-8 rounded-sm px-8 disabled:cursor-not-allowed disabled:opacity-50"
            onClick={() =>
              setSelectableCurrencies((selected) =>
                selected.includes(currency)
                  ? selected.filter((x) => x !== currency)
                  : [...selected, currency]
              )
            }
          >
            <img alt={currency} src={currencyConfig[currency]?.icon} />
            <div className="flex grow flex-col items-start gap-1">
              <div className="text-body uppercase">{currency}</div>
              <div className="text-body-secondary text-xs">
                {currencyConfig[currency]?.unicodeCharacter ?? ""}{" "}
                {currencyConfig[currency]?.name ?? currency}
              </div>
            </div>
            {selectableCurrencies.includes(currency) ? (
              <StarIcon className="stroke-primary fill-primary" />
            ) : (
              <StarIcon />
            )}
          </button>
```

apps/extension/src/ui/atoms/tokenRates.ts
```diff
@@ -38,3 +41,48 @@ export const tokenRatesQuery = selectorFamily({
...
+  effects: [
+    storageEffect(Browser.storage.local, {
+      key: "settings",
+      subKey: "selectableCurrencies",
+      isSet: true,
+    }),
+  ],
+})
+
+export const selectableCurrenciesState = selector<readonly TokenRateCurrency[]>({
```
0xKheops, line 43: Please use only the existing `useSetting` hook for read/write to the settings store
Just need to add 2 properties to SettingsStoreData here : https://github.com/TalismanSociety/talisman/blob/777fbad3d11fa46f9699afa703e371cc0337d707/apps/extension/src/core/domains/app/store.settings.ts#L4
Then can remove these atoms and the storageEffect

0xKheops (dismissed):
Looking great, just requesting a change regarding settings store access, as discussed on discord.

Thinking how "fiat" values were displayed before with 2 decimals and $ symbol felt better, but that's something to discuss during feature crit.

apps/extension/src/ui/atoms/tokenRates.ts
tien: Here we can't just use the hook for plainly getting and setting values though, cause as you can see these atoms have custom logic for get & set value

tien: But I have change the atoms to use the existing settings store instead

0xKheops (commented):
LGTM

letting @chidg doing the real approval as i'm not sure if this must go in 1.19 or not
also might want to wait the feature crit

apps/extension/src/core/domains/app/store.settings.ts
```diff
@@ -12,6 +13,8 @@ export interface SettingsStoreData {
   collapsedFolders?: string[] // persists the collapsed folders in the dashboard account picker
   autoLockTimeout: 0 | 300 | 1800 | 3600
   spiritClanFeatures: boolean
+  selectableCurrency: TokenRateCurrency[]
```
chidg: This is an array, so the property name needs to be plural - `selectableCurrencies`.

apps/extension/src/ui/apps/dashboard/routes/Settings/SettingsPage.tsx
```diff
@@ -127,6 +128,15 @@ export const SettingsPage = () => {
             to={`/settings/language`}
           />
         )}
+        {i18nEnabled && (
```
chidg: I don't think we need to hide this behind the `i18nEnabled` flag

apps/extension/src/ui/atoms/tokenRates.ts
chidg: If we need additional logic on top of the `useSetting` hook, it would be better to create a `useCurrencySetting` hook to expose getting and setting that particular value as we have done with `tokenRatesState`. Still not 100% sure this needs to be in its own atom, also - can't we just put this logic into an ordinary hook which calls `useSetting` and exposes that getting and setting logic? Anyway, if it really does need an atom, let's make a new file for that rather than adding it to `tokenRates`, which is related but slightly different data.

apps/extension/src/ui/domains/Asset/Fiat.tsx
```diff
@@ -42,16 +44,18 @@ const DisplayValue: FC<DisplayValueProps> = ({ amount, currency, noCountUp }) =>
...
   noCountUp = false,
   isBalance = false,
+  hideSymbol = false,
 }: FiatProps) => {
   const { refReveal, isRevealable, isRevealed, isHidden, effectiveNoCountUp } =
     useRevealableBalance(isBalance, noCountUp)
 
   const render = amount !== null && amount !== undefined
 
+  const currency = useRecoilValue(selectedCurrencyState)
```
chidg: Would prefer to see this wrapped in a `useCurrencySetting` hook as we have done with all other recoil state access

apps/extension/src/ui/hooks/useCurrency.tsx
```diff
@@ -0,0 +1,19 @@
+import { selectableCurrenciesState, selectedCurrencyState } from "@ui/atoms"
+import { useRecoilCallback } from "recoil"
+
+export const useToggleCurrency = () =>
```
chidg, line 6: Let's give the file the same name as the hook, since `useCurrency` is a bit general and doesn't describe the function of the hook.

chidg (changes requested):
Good stuff overall, few changes requested

chidg:
`D` doesn't look centered in the circle on the fullscreen portfolio

Screenshot 2023-09-06 at 1 42 51 pm

chidg:
Possible to display the currency code after the value? (so - in this case it would say `$ 211.98 AUD`)
Alternatively display the currency code in a tooltip on hover over the `$`

Screenshot 2023-09-06 at 1 44 54 pm

Inversely, here we do not have the narrow symbol. It would be great to have the same format (`$211.98 AUD`) in all places.
Screenshot 2023-09-06 at 1 46 10 pm

Design thoughts on this @diogoalmeidaribeiro and @0byt ?

apps/extension/src/ui/atoms/tokenRates.ts
tien: Similar to stuff like redux & redux saga, one advantage of using state lib like Recoil is the ability to completely extract business logic from the view/render layer. So separation of concern + in the case of recoil, automatic memoization of selector, so computation will only happen once rather than on every re-render.

apps/extension/src/ui/hooks/useCurrency.tsx
tien: I'm following naming pattern used by others files that exports multiple hooks i.e `useSettings`, `useFeatures`. Right now, this file only exports a single hook, but naming it `useCurrency` allows us to add more hooks later without changing the file name.

apps/extension/src/ui/atoms/tokenRates.ts
tien: How's another file named `currency.ts` sound?

apps/extension/src/ui/domains/Asset/Fiat.tsx
tien: Hmm, not sure whether wrapping every recoil in a hook is the best approach. Because, recoil have multiple access patterns, all with different re-render behaviour (`useRecoilState`, `useRecoilValue`, `useRecoilValueLoadable`, etc).

Though not applicable now, having direct access to recoil state allows concurrent suspense pattern like this:

```tsx
const [asyncValueA, asyncValueB, asyncValueC] = useRecoilValue(waitForAll([asyncStateA, asyncStateB, asyncStateC]));
```

tien:
> Possible to display the currency code after the value? (so - in this case it would say `$ 211.98 AUD`) Alternatively display the currency code in a tooltip on hover over the `$`
> 
> Screenshot 2023-09-06 at 1 44 54 pm
> 
> Inversely, here we do not have the narrow symbol. It would be great to have the same format (`$211.98 AUD`) in all places. Screenshot 2023-09-06 at 1 46 10 pm
> 
> Design thoughts on this @diogoalmeidaribeiro and @0byt ?

The format is controlled by [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) so not sure whether this pattern `$211.98 AUD` is possible

0byt:
Yeah, ideally the AUD is after the number, but if not possible, then I'm not picky

apps/extension/src/ui/domains/Asset/Fiat.tsx
chidg: That's good to keep in mind for the future, but in this PR there's only a single access pattern used for this atom, and it's used in quite a few places, so it definitely makes sense to consolidate the behaviour to a reusable hook

apps/extension/src/ui/hooks/useCurrency.tsx
chidg: There are some exceptions, but the main pattern in the codebase by far is to have the file name the same as the main exported hook function.

apps/extension/src/ui/domains/Asset/Fiat.tsx
chidg: See `useToken` and `useTokenRates` for good examples of the type of hook I think we need in this case.

apps/extension/src/ui/hooks/useCurrency.tsx
tien: With the addition of `useSelectedCurrency` this file name now make sense

chidg (approved):
LGTM

PR #1026, opened 2023-08-21
fix: incorrect cached/stale balances for evm accounts

### LINKED ISSUE #1019: Incorrect `stale`/`cache` status on DOT & KSM
Potentially two unrelated issues:
- [cached balances (pulsing balances bg)](https://discord.com/channels/969497836410507274/969971058918699110/1140893664655200286)
- stale balances (warning icon shown next to balances)

---

- `cache` status (pulsing balance bg) only shows when viewing `total portfolio`
- In the balances store, it looks as though all evm-type accounts have perpetually `stale` balances

alecdwm:
[Fixes the sneaky bug introduced by this commit](https://github.com/TalismanSociety/talisman/commit/d2ccdafba8d97023e58f2234dfef3d153de0b2e1#diff-7253ed9dc13e7a9171ac4fbef647aa74c4886356e9bc087151534000f61b90c5R491-R511).

Because we were no longer making unnecessary state calls (System.Account `address` for evm accounts on non-evm chains) the store was left with a bunch of `Balance` rows which would go stale/remain forever cached.

This PR patches the problem by continuing to make those calls for now.

The better fix would be to delete the rows from the DB, but I don't want to set up any fancy migrations sort of code for this area yet.

There are other limitations on the current balances db structure which will need to be addressed soon anyway. So, I don't think it's worth trying to polish up what's there.

Specifically - the current structure doesn't allow us to have different statuses (e.g. `cached`, `live`) for a token whose balance comes from multiple pallets - e.g. nomination pools:
- DOT (System.Account)
- DOT (nominationPools)

When we fix that we will implicitly ditch the current cache anyway :)

### DIFF
--- a/.changeset/angry-bottles-look.md
+++ b/.changeset/angry-bottles-look.md
@@ -1,0 +1,5 @@
+---
+"@talismn/balances-substrate-native": patch
+---
+
+fix: incorrect cached/stale balances for evm accounts
--- a/packages/balances-substrate-native/src/SubstrateNativeModule.ts
+++ b/packages/balances-substrate-native/src/SubstrateNativeModule.ts
@@ -496,18 +496,33 @@ async function buildQueries(
           "account",
           decodeAnyAddress(address)
         )
-        const stateKey = hasMetadataV14
-          ? storageHelper.stateKey
-          : (() => {
-              const addressBytes = decodeAnyAddress(address)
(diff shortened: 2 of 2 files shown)


alecdwm:
Closes https://github.com/TalismanSociety/talisman/issues/1019

0xKheops (approved):
LGTM

PR #1027, opened 2023-08-21
fix: remove old tokenRates for tokens which have no coingeckoId anymore

alecdwm:
When we remove coingeckoIds from chaindata, the wallet just keeps the last-fetched rate for the token instead of removing it from the tokenRatesDb.

This PR fixes that.

### DIFF
--- a/.changeset/gold-camels-brush.md
+++ b/.changeset/gold-camels-brush.md
@@ -1,0 +1,5 @@
+---
+"@talismn/balances-react": patch
+---
+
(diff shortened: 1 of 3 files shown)


alecdwm:
For testing you can check that the rate for `qDOT` goes from `$X.XX` to `-` in your portfolio.

Fresh installs will just show `-` by default, but existing installs will have the rate still.

0xKheops (approved):
LGTM

PR #1029, opened 2023-08-22
fix: checkPassword if pw not hashed (v1)

0xKheops:
allows users that didn't migrate to password V2 yet to go through the migration wizard.


chidg (approved):
LGTM

PR #1030, opened 2023-08-22
fix: update caniuse-lite

alecdwm:
Version `1.0.30001409` is a year old

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11351,17 +11351,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"caniuse-lite@npm:^1.0.30001400, caniuse-lite@npm:^1.0.30001407":
-  version: 1.0.30001409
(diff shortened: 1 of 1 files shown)


0xKheops (commented):
agree it feels necessary, but unsure about the mechanism that was used to update the yarn lock like this

what about bumping autoprefixer ? 
it is one of our dependencies, being a TW's prerequisite, it's in charge of css generation with the correct browser css prefixes.
if upgrading it from 10.4.12 to 10.4.15, it would upgrade caniuse-lite to  1.0.30001520

0xKheops:
just noticed the message when building, was it a `npx update-browserslist-db@latest` ?

alecdwm:
It was!

0xKheops (approved):
LGTM!

PR #1033, opened 2023-08-23
chore: cleanup hook

0xKheops:
noticed the hook wasn't cleaned up
wanted to check something about easings which led me to do it, so here's the PR :)

(this targets your PR)


chidg (approved):
Even better, didn't notice this part!
LGTM

PR #1036, opened 2023-08-24
feat: show mnemonic before creating account

0xKheops:


apps/extension/src/core/domains/accounts/handler.ts
```diff
@@ -45,63 +37,50 @@ import { SOURCES } from "../mnemonics/store"
...
-        name,
-        origin: AccountTypes.TALISMAN,
-        ...accountMeta,
-      })
-      talismanAnalytics.capture("account create", { type, method: "parent" })
-      return pair.address
-    } else {
-      const { val: derivationPath, err } = getNextDerivationPathForMnemonic(seed, type)
-      if (err) throw new Error(derivationPath)
+    const mnemonicResult = await this.stores.seedPhrase.getSeed(derivedMnemonicId, password)
```
chidg: It seems like this should go in the `if ("mnemonicId" in options)` block above, since in the other case, we know whether or not we are able to successfully store the mnemonic

apps/extension/src/core/domains/accounts/handler.ts
```diff
@@ -124,30 +103,41 @@ export default class AccountsHandler extends ExtensionHandler {
...
+          true
+        )
+        if (result.ok) {
+          meta.origin = AccountTypes.SEED // find a better name
+          meta.derivedMnemonicId = result.val
+          meta.derivationPath = derivationPath
+        } else throw new Error("Failed to store mnemonic", { cause: result.val })
+      }
+    } else {
+      meta.origin = AccountTypes.SEED // TODO "LOCAL"
```
chidg, line 140: In the case that `mnemonicValidate` is false, we are assuming that suri is a valid private key. Should we also validate that to ensure it isn't just garbage?

apps/extension/src/core/domains/accounts/helpers.ts
```diff
@@ -155,49 +156,36 @@ export const getPrimaryAccount = (storedSeedOnly = false) => {
...
   string,
   "Unable to get next derivation path" | "Reached maximum number of derived accounts"
 > => {
-  const allAccounts = keyring.getAccounts().filter(({ meta }) => meta && meta.type === type)
-  let accountIndex
-  let derivedAddress: string | null = null
+  const allAccounts = keyring.getAccounts()
   try {
+    // for substrate check empty derivation path first
+    if (type !== "ethereum") {
```
chidg, line 170: Good to consider that we may have other non-ethereum types in future 👍🏼

apps/extension/src/ui/domains/Account/AccountCreate/AccountAddDerivedForm.tsx
```diff
@@ -53,8 +80,40 @@ export const AccountAddDerivedForm = ({ onSuccess }: { onSuccess: (address: stri
...
+    () => [
+      ...mnemonics
+        .map((m) => ({
+          label: m.name,
+          value: m.id,
+          accounts: allAccounts.filter((a) => a.derivedMnemonicId === m.id),
+        }))
+        .sort((a, b) => a.label.localeCompare(b.label)),
+      GENERATE_MNEMONIC_OPTION,
+    ],
```
chidg:
```suggestion
    () => {
      const accountsByMnemonic = allAccounts.reduce((result, acc) => {
        if (!acc.derivedMnemonicId in result) result[acc.derivedMnemonicId] = []
        result[acc.derivedMnemonicId].append(acc)
        return result
      }, {} as Record<string, AccountJsonAny[]>)
      return [
      ...mnemonics
        .map((m) => ({
          label: m.name,
          value: m.id,
          accounts: accountsByMnemonic[m.id] || [],
        }))
        .sort((a, b) => a.label.localeCompare(b.label)),
      GENERATE_MNEMONIC_OPTION,
    ]},
```

chidg (changes requested):
Small changes suggested

apps/extension/src/core/domains/accounts/handler.ts
0xKheops: I think no need, if it was invalid the keyring.addUri call could break

0xKheops: image

chidg (approved):
LGTM

PR #1037, opened 2023-08-26
fix: extrinsic payload parsing

0xKheops:
Found out that extrinsic payloads sent from dapps may not always use the same format.
Some dapps (at least the ones built with CAPI) craft payloads where the specVersion can be a number instead of hex string.
This displayed a warning in Talisman's signing popup, and probably prevented QR codes from being generated.

This PR fixes associated issues by decoding the payload using TypeRegistry before reading the specVersion.


0xKheops:
nevermind

PR #1047, opened 2023-08-30
More onboarding improvements

chidg:
Still todo:
- [x] Add account (dashboard) : Main component being vertically centered creates a layout shift when selecting any option. should be docked at the top like other screens.
- [ ] Import hardware devices : wording should be changed from "Import" to "Connect" (or "Attach?") in all page titles, for consistency with add account first screen WONTFIX


apps/extension/src/ui/apps/popup/components/TotalFiatBalance.tsx
```diff
@@ -44,30 +42,24 @@ export const TotalFiatBalance = ({ className, mouseOver }: Props) => {
...
+        </button>
       </div>
       <Fiat
         className="font-surtExpanded text-lg"
         amount={balances.sum.fiat("usd").total}
         currency="usd"
         isBalance
       />
-      {hasAccounts && <TopActions />}
+      <TopActions />
```
0xKheops, line 62: might want to hide these if no accounts

apps/extension/src/ui/apps/popup/pages/Login.tsx
```diff
@@ -171,7 +171,7 @@ const useDevModeAutologin = ({
   useLayoutEffect(() => {
     if (process.env.NODE_ENV === "production") return
     if (!process.env.PASSWORD) return
-    setValue("password", process.env.PASSWORD)
+    // setValue("password", process.env.PASSWORD)
```
0xKheops: can't live without this :)

apps/extension/src/ui/apps/onboard/routes/Password.tsx
```diff
@@ -103,9 +101,9 @@ export const PasswordPage = () => {
         name: "Submit",
         action: "Choose password continue button",
       })
-      navigate(isResettingWallet ? "/account" : `/privacy`)
+      navigateNext()
```
0xKheops, line 101: need to revert this line change, as navigateNext is memoised, this will run with isLoggedIn === FALSE

otherwise we're redirected to dashboard right after setting password, skipping privacy and account creation screens

0xKheops (changes requested):
Found a problem on Password.tsx, see associated comment 
Also reonboarding after reset password doesn't work, I get redirected to onboarding home page when trying to progress through the wizard

apps/extension/src/ui/apps/popup/components/TotalFiatBalance.tsx
chidg: This entire component will no longer be shown if no accounts, so it isn't needed here

apps/extension/src/ui/apps/onboard/routes/Password.tsx
chidg: In this case it wasn't because of memoisation, but simply because before the password is set, `isLoggedIn` is FALSE 🤦‍♂️.

apps/extension/src/ui/apps/popup/components/TotalFiatBalance.tsx
0xKheops: it did appear for me when i tested, with no account

0xKheops (approved):
LGTM !

PR #1048, opened 2023-08-30
feat: custom derivation path

0xKheops:
- [x] import account form
- [ ] new account form
- [ ] dedicated page for PK import ? (makes no sense to show derivation path dropdown when importing a PK)


0xKheops:
replaced by 1049

PR #1049, opened 2023-08-30
feat: custom derivation path - 2nd attempt

### LINKED ISSUE #989: Custom Derivation Path

0xKheops:
- [x] import account form
- [x] new account form


apps/extension/src/ui/apps/onboard/routes/AddAccount/AccountAddDerivedPage.tsx
```diff
@@ -1,10 +1,15 @@
...
-    title="Create a new account"
-    render={(onSuccess) => <AccountAddDerivedForm onSuccess={onSuccess} />}
-  />
-)
+export const AccountAddDerivedPage = () => {
+  const { t } = useTranslation("admin")
+
+  return (
+    <AccountAddWrapper
+      title={t("Create a new account")}
```
chidg, line 11: The `t` function is called on the props given inside `AccountAddWrapper`. I wasn't entirely sure if this would work though and hadn't tested it yet. If it doesn't work, we should also update `AccountAddWrapper` to remove the `t` call there.

chidg:
A few general UX observations:
- When I go to derive a 2nd account from an existing stored mnemonic, and choose custom derivation path, the placeholder displays the first derivation path, and the error message "Address already exists" appears. Can we delay validation until the input has been touched?
- Unchecking the 'custom derivation path' input does not clear this error message, and the error is still not cleared when I re-check it and enter something in the input in some cases
- There is a layout issue on the form, notice the top of the checkbox has been clipped by the element above
Screenshot 2023-09-05 at 11 35 18 am

- Maybe it would be preferable to have the derivation path as a default value on the input rather than a placeholder - then it could be edited in place rather than entered from scratch as it is on the import from recovery phrase screen.
- Entering an invalid derivation path is not caught by the form
Screenshot 2023-09-05 at 11 43 20 am

- For the import recovery phrase screen, the form is valid when custom derivation derivation path is chosen for an EVM account, but the derivation path input is empty. Not sure if this is intended or not?

0xKheops:
> For the import recovery phrase screen, the form is valid when custom derivation derivation path is chosen for an EVM account, but the derivation path input is empty. Not sure if this is intended or not?

Yes empty derivation path is valid

chidg (approved):
LGTM!

PR #1052, opened 2023-09-03
fix: inject error on unauthorised sites

0xKheops:
Fixes an error making extension unable to inject on sites that are not yet authorised. 
`connectWatchedAccounts` property was being set on an object that doesn't exist, resulting in an unexpected error

Solution : 
- removed the `connectWatchedAccounts` property from AuthorisedSite object as it doesn't need to be in the store
- instead call `isTalismanHostname(site.url)` when needed at runtime


alecdwm (approved):
Awesome, LGTM! 🎉

PR #1085, opened 2023-09-21
fix: onboarding flicker

### LINKED ISSUE #1082: Flickering in onboarding UI (first seen with MysticalBackground V3)
**Describe the bug**
On some systems, the onboarding background can cause some hectic flickering.

**To Reproduce**
1. Install Talisman on one of the setups listed under *Environments*.

**Screenshots**
Screen Recording 2023-09-19 at 11 06 56

Screen Recording 2023-09-19 at 11 06 56

**Environment**
Confirmed to occur on these systems:
- Chrome on macOS (arm64) - 3x separate computers so far
- Arc on macOS (arm64) - 2x separate computers so far

Not seen on these systems (yet):
- Firefox on macOS (arm64)
  - But slow performance **is** observed
- Chrome on macOS (x86_64)
- Chrome on Linux (x86_64)

alecdwm:
Fixes https://github.com/TalismanSociety/talisman/issues/1082

**Works on My Machine™**  
(We should definitely do some testing on different systems)

Screen Recording 2023-09-21 at 14 10 26

I left the WIP commits in for this one, so when reviewing it would be better to look at the full diff of the PR against dev, instead of at each individual commit.


packages/talisman-ui/src/components/MysticalBackgroundV3/MysticalPhysicsV3.ts
```diff
@@ -13,7 +13,7 @@ export type MysticalPhysicsV3 = {
 
 export const MYSTICAL_PHYSICS_V3: MysticalPhysicsV3 = {
   artifacts: 3,
-  blur: 24,
+  blur: 0,
```
alecdwm, line 16: This actually didn't work before!
I've fixed it so that setting `blur` actually does something - but then I've set the default config to not use any blur - since we weren't using it and it looks good without it anyway (and it costs a bit to render it)

packages/talisman-ui/src/components/MysticalBackgroundV3/MysticalBackgroundV3.tsx
```diff
@@ -1,111 +1,110 @@
...
+  const mergedConfig = useMemo(() => ({ ...MYSTICAL_PHYSICS_V3, ...(config ?? {}) }), [config])
 
-const CelestialArtifacts: FC<{
-  size: ParentSize
-  config: MysticalPhysicsV3
-}> = ({ size, config }) => {
-  const artifactKeys = useMemo(() => Array.from(Array(config.artifacts).keys()), [config.artifacts])
+  const style = useMemo<CSSProperties>(
+    () => ({
+      filter: mergedConfig.blur > 0 ? `blur(${mergedConfig.blur}px)` : undefined,
```
alecdwm, line 20: This fixes the blur (was previously `transition: blur()` which is invalid)

packages/talisman-ui/src/components/MysticalBackgroundV3/MysticalBackgroundV3.tsx
```diff
@@ -1,111 +1,110 @@
...
-    useEffect(() => {
-      refInitialized.current = false
-    }, [])
-
-    const [style1, style2] = useMemo<[CSSProperties, CSSProperties]>(
-      () => [
-        {
-          stopColor: artifact.color,
-          stopOpacity: 0.9,
-          transitionProperty: "stop-color",
```
alecdwm, line 33: The primary source of our flickering was the repainting required to transition the `stop-color` on these gradients.

packages/talisman-ui/src/components/MysticalBackgroundV3/MysticalBackgroundV3.tsx
```diff
@@ -1,111 +1,110 @@
...
-CelestialArtifact.displayName = "CelestialArtifact"
+export const MysticalBackgroundV3 = ({
+  className,
+  config,
+}: {
+  config?: Partial<MysticalPhysicsV3>
+  className?: string
+}) => {
+  const [refSize, size] = useMeasure<HTMLDivElement>()
+  const mergedConfig = useMemo(() => ({ ...MYSTICAL_PHYSICS_V3, ...(config ?? {}) }), [config])
```
alecdwm, line 16: Now memoized!

packages/talisman-ui/src/components/MysticalBackgroundV3/useCelestialArtifact.ts
```diff
@@ -1,95 +1,175 @@
...
-      transitionDuration: `${duration}ms`,
-      transitionDelay: "100ms", // prevents flickering on FF
-      transitionTimingFunction: initialized ? "ease-in-out" : "ease-out",
-      opacity: Number(
-        (config.opacityMin + Math.random() * (config.opacityMax - config.opacityMin)).toFixed(2)
-      ),
-    },
-  }
+  /** Is used to ensure a consistent z-order of artifacts, i.e. to avoid any flickering from artifacts competing to be on top */
+  artifactIndex: number
```
alecdwm, line 14: This manual z-order handling isn't *strictly* necessary, but I added it for good measure since we're doing 3d translations anyway (supposedly better performance because they're isolated from the rest of the page and sent to the GPU).

packages/talisman-ui/src/components/MysticalBackgroundV3/useCelestialArtifact.ts
```diff
@@ -1,95 +1,175 @@
...
+  const maxSize = Math.min(parentSize.width, parentSize.height) * 2
+  const width = Math.round(maxSize * randBetween(config.radiusMin, config.radiusMax))
+  const height = Math.round(width * (1 - config.ellipsisRatio * Math.random()))
+
+  const x = Math.round(Math.random() * parentSize.width)
+  const y = Math.round(Math.random() * parentSize.height - Math.max(width / 2, height / 2) / 2)
+
+  const color = forceColor ?? Color.hsv(Math.random() * 360, 100, 100).hex()
+  const opacity = randBetween(config.opacityMin, config.opacityMax)
+  const background = artifactBg(color, opacity)
```
alecdwm, line 106: We can't set the `opacity` on the artifact itself, because that's the prop we're using to transition between the two artifact colors!

So - I've just baked the intended opacity into the `background` gradient.

packages/talisman-ui/src/components/MysticalBackgroundV3/useCelestialArtifact.ts
```diff
@@ -1,95 +1,175 @@
...
+  //
+  // However, on firefox the radial-gradient background is computed *before* scaling the element.
+  // So if we use `1px`, we'll have a radial-gradient scaled to `500px` in screen size but with the resolution of just one pixel!
+  //
+  // This would *not* be a very nice gradient to look at :^)
+  //
+  // So to fix this, we'll scale this value up a few orders of magnitude, to `1000px`.
+  // And then for our 500x500px circle we need to use `scale3d(0.5, 0.5, 1)`.
+  // Because `500` (the size we want) divided by `1000` (our base size, or backgroundResolution) is `0.5`.
+  const backgroundResolution = 1000 // in pixels squared
```
alecdwm, line 125: TLDR:
We want to scale big thing down to smaller size, not small thing (1x1px) up to bigger size.

It doesn't matter either way in chrome, but in firefox it breaks the whole effect when you go from small to big.

apps/extension/src/ui/apps/popup/components/LoginBackground.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export type LoginBackgroundProps = {
+  className?: string
+  colors: [string, string]
+}
+export const LoginBackground = ({ className, colors }: LoginBackgroundProps) => {
+  const config = useMemo(() => ({ ...BG_CONFIG, colors }), [colors])
+
+  return <MysticalBackground className={className} config={config} />
+}
```
alecdwm, line 19: Made this consistent with `OnboardingBackground` and the popup's `AllAccountsHeaderBackground`

apps/extension/src/ui/apps/popup/pages/Login.tsx
```diff
@@ -36,14 +36,7 @@ const INPUT_CONTAINER_PROPS = { className: "bg-white/10" }
...
 
-  return (
-    <LoginBackground
-      width={400}
-      height={600}
-      colors={colors}
-      className="absolute left-0 top-0 m-0 block h-full w-full overflow-hidden "
-    />
-  )
+  return <LoginBackground className="absolute left-0 top-0 h-full w-full" colors={colors} />
```
alecdwm, line 39: The removed props & tailwind classes were redundant

chidg (approved):
LGTM
Works on my machine in chrome and FF. Slightly jerky movement in FF but possibly due to lots of other tabs open. Let's confirm with some more eyes on it and then good to go.

PR #1087, opened 2023-09-22
fix: use simple localize api rather than cdn to get correct language list

chidg:
Using the Simple Localize CDN served cached versions of the active language list. This PR uses the API rather than the CDN to get up to date languages.


alecdwm (approved):
LGTM!

PR #1089, opened 2023-09-26
feat: asset price display

tien:


apps/extension/src/ui/domains/Portfolio/useTokenBalancesSummary.ts
```diff
@@ -155,5 +155,11 @@ export const useTokenBalancesSummary = (balances: Balances) => {
     return summary
   }, [currency, tokenBalanceRates, tokenBalances.count, tokenBalances.each])
 
-  return { token, summary, tokenBalances, tokenBalanceRates }
+  return {
+    token,
+    rates: token?.id !== undefined ? tokenBalanceRates[token?.id] : undefined,
```
chidg: Rather than exporting this `rates` object here and then having to import the `useSelectedCurrency` hook into every component where we want to get the rate for the selected currency, I think it would be preferable to do it all in one place and return the rate for the selected currency from this hook, since we already have `currency` here, and there is no reason why we need the other rates in the UI.

chidg (changes requested):
See comment, small change requested

chidg:
Just a reminder on this one @tien

chidg (approved):
LGTM

PR #1091, opened 2023-09-26
Minor updates pre-1.19

chidg:


apps/extension/src/ui/domains/SendFunds/useNetworkDetails.ts
```diff
@@ -0,0 +1,32 @@
+import { Chain, CustomChain, CustomEvmNetwork, EvmNetwork } from "@talismn/chaindata-provider"
+import { useCallback, useMemo } from "react"
+import { useTranslation } from "react-i18next"
+
+import { useSendFunds } from "./useSendFunds"
+
+export const useFormatNetworkName = () => {
+  const [t] = useTranslation()
+
+  return useCallback(
```
alecdwm, line 20: It's maybe a little unclear what's happening here without context, could do with a comment like `We want to add a suffix to chain names for chains with both a substrate and an evm RPC`

alecdwm (dismissed):
Looks good to me!

apps/extension/src/ui/hooks/ledger/common.ts
```diff
@@ -49,11 +49,12 @@ export const ledgerNetworks = [
...
   //   genesisHash: "0x68d56f15f85d3136970ec16946040bc1752654e906147f7e43e9d539d7c3de2f",
   // },
-  {
-    name: "centrifuge",
-    genesisHash: "0xb3db41421702df9a7fcac62b53ffeac85f7853cc4e689e0b93aeb3db18c09d82",
-    label: "Centrifuge",
-  },
+  // Removed because no longer maintained https://support.ledger.com/hc/en-us/articles/11158925739677-Centrifuge-CFG-
+  // {
+  //   name: "centrifuge",
```
0xKheops, line 54: bit worried about what happens for users that already created an account for centrifuge.
might want to add a "legacy" property on entries of this array to prevent account creation, but keep existing accounts behave as before ?

apps/extension/src/ui/domains/Asset/TokenPicker.tsx
```diff
@@ -189,6 +190,7 @@ const TokensList: FC<TokensListProps> = ({
   const { evmNetworksMap } = useEvmNetworks(useTestnets)
   const { tokens: allTokens } = useTokens(useTestnets)
   const tokenRatesMap = useTokenRatesMap()
+  const formatNetworkName = useFormatNetworkName()
```
0xKheops, line 193: having a hook returning just a callback feels weird, wdyt about making chain & evmNetwork arguments of the hook, and have the hook return the formatted label directly ?

0xKheops (commented):
looking good, just a couple questions before approving

apps/extension/src/ui/domains/Asset/TokenPicker.tsx
chidg: It's a valid pattern. The function needs to be used in places where you can't directly call a hook, like inside another hook in this component.

apps/extension/src/ui/hooks/ledger/common.ts
chidg: Discussed in Discord - result of doing this is that existing accounts can still be viewed and balances are fetched, but transactions are not able to be signed with them, and new accounts are not able to be created for the network.

0xKheops (approved):
LGTM !

alecdwm (approved):
LGTM2!

PR #1094, opened 2023-09-28
Ensure polkadot vault verifier cert mnemonic is stored as confirmed during migration

chidg:
Fixes an error in 1.19.0 RC1 where during migrations for the mnemonic store, an imported polkadot vault verifier certificate mnemonic would have `confirmed: false` when migrated to the new mnemonics store.


0xKheops (approved):
LGTM

Issue #1076, opened 2023-09-12
Decephalopoded Chaindata


alecdwm:
##### Latest Summary before we can merge:

- [x] Balance modules TypeRegistry -> subshape
- [x] Make new tokens stuff prod ready
  - Check that chain and token icons are correct
  - Need to fetch tokens list and miniMetadatas from the new chaindata cache
  - MiniMetadataUpdater needs to be converted from "working" to "reliable" and fast

---

- [ ] Throw some errors deep in the code and find out where they surface, to make sure that we're handling them appropriately
- [x] Update balance module docs

---

- [ ] (Alec/Kheops) make sure that token toggles are working

##### Wallet repo:
- [x] Consolidate balance module libs
- [ ] Consolidate other libs
- [x] Refactor balance modules to use `@talismn/scale` instead of `@talismn/mutate-metadata`
- [x] Refactor `@talismn/chaindata-provider` to pull from github pages endpoint
- [x] Add module (probably internal to `@talismn/balances`) which stores and updates miniMetadatas
  - To start don't bother putting it in a web worker, come back to that (but when designing it keep in mind that it should eventually go in a web worker)
- [x] Update hooks to use module (either run in react useEffect or in spawned web worker)
- [ ] Write new wallet balances integration (balances.store.ts) to use spawned web worker
- [ ] Use balances-react hooks to retrieve balances directly from store
- [ ] ?
- [x] Run metadata stuff in chaindata so that the wallet can just fetch latest status from there (and cache it as a JSON along with chains.json and evm-networks.json) and will only need to run the miniMetadataUpdater on custom / out of date networks
  - [x] This process should use the chainsExtraCache so that it doesn't do unnecessary work
- [ ] ?
- [ ] Update docs & READMEs

##### Chaindata repo:
- [x] Add an action which reads the config files from the repo, derives up to date values for each of the queryable chain properties and then publishes the results to a github pages branch
  - While building, consider whether the action should update every chain at once or if it should update individual chains/batches of chains at a time
- [x] Add a trigger to run the action on (a) a timer, and (b) on changes to the main branch
- [x] Remove `tokens.json` and merge into the `balanceModuleConfigs` chain key
  - Perhaps also rename the `balanceModuleConfigs` key while we're at it

#### Notes/Scope

- Consolidate all balance libs into just `@talismn/balances` and `@talismn/balances-react`
- Consolidate `@talismn/chaindata-provider-extension` into `@talismn/chaindata-provider`
[...]

PR #1101, opened 2023-10-01
refactor: distributed metadata

### LINKED ISSUE #1152: Wallet incorrectly shows returned crowdloan funds as still locked
- Crowdloan-locked DOT/KSM is fetched inside SubNativeModule.ts
- The chain definitely clears these amounts from state when the DOT/KSM is refunded
- However, we have had many reports that the wallet is still showing the refunded amounts

Possible causes to investigate:
1. Wallet can't subscribe to these amounts, perhaps it is also not refreshing them often enough.
2. Perhaps when the amount goes from `N DOT/KSM` to `0 DOT/KSM`, we're treating `0` as a null value and therefore retaining the `N` value in indexeddb.

alecdwm:
Scoped @ https://github.com/TalismanSociety/talisman/issues/1076


packages/balances/src/MiniMetadataUpdater.ts
```diff
@@ -0,0 +1,261 @@
...
+ *
+ * By utilizing the wonderful [subshape](https://github.com/paritytech/subshape) library,
+ * we can trim the chain metadata down so that it only includes the types we need for balance fetching.
+ *
+ * Each balance module has a function to do just that, `BalanceModule::fetchSubstrateChainMeta`.
+ *
+ * But, we only want to run this operation when necessary.
+ *
+ * The purpose of this class, `MiniMetadataUpdater`, is to maintain a local cache of
+ * trimmed-down metadatas, which we'll refer to as `MiniMetadatas`.
```
chidg, line 42: 😍

packages/balances/src/MiniMetadataUpdater.ts
```diff
@@ -0,0 +1,261 @@
...
+              ),
+          ],
+        ]
+      })
+    )
+
+    const statusesByChain = new Map<string, MiniMetadataStatus>(
+      Array.from(wantedIdsByChain.entries()).map(([chainId, wantedIds]) => [
+        chainId,
+        wantedIds.every((wantedId) => ids.includes(wantedId)) ? "good" : "none",
```
chidg, line 110: Slight nitpick, but can we have these statuses defined as constants or an enum?

chidg (commented):
Small comments

packages/balances/src/MiniMetadataUpdater.ts
chidg: (appreciate the detailed docstring)

apps/extension/src/ui/apps/dashboard/routes/Portfolio/PortfolioAssets.tsx
```diff
@@ -43,35 +44,67 @@ const FullscreenPortfolioAssets = ({ balances }: { balances: Balances }) => {
...
   return (
-    <div className="flex w-full flex-col">
-      {hasAccounts === false && (
-        <div className="mt-[3.8rem] flex grow items-center justify-center">
-          <NoAccountsFullscreen />
-        </div>
-      )}
-      {hasAccounts && <FullscreenPortfolioAssets balances={balancesToDisplay} />}
+    <div className="text-body-secondary mt-72 flex flex-col items-center justify-center gap-8 text-center">
+      <div>{t("Enable some networks to display your assets")}</div>
```
chidg, line 64: When there is a selected account active, this message doesn't quite make sense. In that case we should say "No networks are enabled for this account. Enable some {{ Polkadot || Ethereum }} networks to display your assets."

apps/extension/src/ui/hooks/useChains.tsx
```diff
@@ -7,27 +9,35 @@ import {
 import { useMemo } from "react"
 import { useRecoilValue } from "recoil"
 
-export const useChains = (withTestnets: boolean) => {
-  // // keep db data ption("chains")
+type ChainsFilter = "all" | "enabledWithTestnets" | "enabledWithoutTestnets"
```
chidg: Prefer to use Enums or constants for values like this
Although, since we now have two states (all/enabled and testnets on/off) perhaps an options object would be more suitable?

chidg (commented):
Small comments, mostly very good

apps/extension/src/ui/hooks/useChains.tsx
0xKheops: done in asset discovery pr

alecdwm:
Marking this a draft until we merge assets discovery back into it

chidg (approved):
LGTM 🪬

PR #1108, opened 2023-10-05
fix: faster load of token picker if many accounts

0xKheops:
Optimizes the send funds token picker loading time.

- reuses the  result of `balances.each`
- creates new `Balances` object after filtering positive balances

=> 50% perf gain with 60 accounts


alecdwm (approved):
LFG 🔥

PR #1112, opened 2023-10-06
fix: send funds input amount as fiat

0xKheops:
Uses user defined currency's convertion rate when typing an amount as fiat


alecdwm (approved):
🔥

PR #1123, opened 2023-10-12
fix: send funds address format

0xKheops:
fixes address format in send funds wizard : 
- use selected token's chain's prefix for formatting addresses in account pickers & pills
- for the recipient do not display token's chain's format if address has been manually input, except in the confirm form address tooltip (allows visual confirmation that the address is the intended one)
- in the confirm form address tooltip, do not show original format for talisman accounts, only display the token's chain's address format


chidg (approved):
LGTM

PR #1124, opened 2023-10-12
fix: error message if insufficient tokens

### LINKED ISSUE #1098: Improve send funds insufficient gas error message
Screenshot shows error message when attempting to send funds on Ethereum mainnet from an account with 0 Eth. The problem is that there is insufficient gas to pay for the transaction, but the error message is generic and unhelpful. 

Screenshot 2023-09-29 at 12 01 10 pm

0xKheops:
- Fixes #1098


chidg (approved):
LGTM

PR #1130, opened 2023-10-25
chore: viem

0xKheops:
:fire:


apps/extension/src/core/domains/ethereum/handler.extension.ts
```diff
@@ -65,7 +66,7 @@ export class EthHandler extends ExtensionHandler {
           method,
           hostName: ok ? host : null,
           dapp: queued.url,
-          chain: chainId,
+          chain: Number(ethChainId),
```
chidg, line 69: I notice in other places in this file we send a Posthog event without coercing `ethChainId` to a number, we should be consistent about that because it could be a hexstring in some places

apps/extension/src/core/domains/ethereum/types.ts
```diff
@@ -16,42 +34,101 @@ export type {
...
+    Extract<TRpcSchema[number], { Method: K }>["ReturnType"]
+  ]
+}
+
+// type EthRequestSignaturesMapBetter<TRpcSchema extends RpcSchema> = {
+//   [K in TRpcSchema[number]["Method"]]: {
+//     parameters: Extract<TRpcSchema[number], { Method: K }>["Parameters"]
+//     returnType: Extract<TRpcSchema[number], { Method: K }>["ReturnType"]
+//   }
+// }
```
chidg: want to remove this?

chidg (commented):
LGTM
One small issue re: analytics data.

apps/extension/src/core/domains/ethereum/handler.extension.ts
0xKheops: good catch, should be all numbers now

chidg (approved):
LGTM

PR #1138, opened 2023-11-02
Refactor custom network forms

chidg:
Fixes validation race condition, also refactors substrate form to make it more legible


alecdwm (dismissed):
LGTM!!!

apps/extension/src/core/domains/chains/types.ts
```diff
@@ -33,7 +33,7 @@ export type RequestUpsertCustomChain = {
   nativeTokenLogoUrl: string | null
   accountFormat: string | null
   subscanUrl: string | null
-  rpcs: { url: string }[]
+  rpcs: { url: string; genesisHash?: SignerPayloadGenesisHash }[]
```
0xKheops: not sure if we should do this, this defines the argument for backend call which doesn't need this property

alecdwm: Yeah this stood out to me too, but I couldn't think of a better place to put the genesisHash instead

chidg: ok, in that case I think we should separate data needed for the form itself from data needed in the backend and run it through a transform. Sound ok?

0xKheops: sounds great

0xKheops (approved):
LGTM !

Issue #1067, opened 2023-09-11
Support EVM L2


0byt:

0byt:
Target chains:
* Base / Optimism, etc (OP Stack)
* Arbitrum / Arbitrum Nova
* Polygon (?)
* Avail (?)
* Research other possible chains that we need to support

Current State:
* It works these days because we take some margin on our computation (Gas limit with 2% margin)
* For Optimism, currently base fee doesn't work 
* Currently ignoring cost of transaction of L1

Output:
- [ ] UI: Need to show L1/L2 fee, show total to user
- [ ] Wallet: Compute cost of L1 fee
- [ ] Review Send Max feature to take these fees into account
- [ ] Setup infra to be able to scale to other chains later

0xKheops:
Implemented in #1146

PR #1140, opened 2023-11-07
feat: OP stack layer 1 fees

0xKheops:


chidg:
Looks like this needs to be brought up to date after changes upstream. LGTM when that's done!

0xKheops:
Replaced by #1067

PR #1142, opened 2023-11-07
fix: prevent useOnChainId unnecessary refreshes

0xKheops:
Prevents useOnChainId(address) from refetching data each time it's called

test case : navigate back & forth between popup home page and one account's page


chidg (approved):
LGTM

PR #1158, opened 2023-11-23
Open send funds from portfolio asset page with token search prefilled with symbol

chidg:
Uses the token symbol as the initial search input for the token picker when 'send funds' is selected from the button at the top of the portfolio asset detail page.


0xKheops (approved):

0xKheops (approved):
LGTM!

PR #1159, opened 2023-11-24
Analytics Improvements

chidg:


apps/extension/src/core/domains/accounts/handler.ts
```diff
@@ -203,17 +211,17 @@ export default class AccountsHandler extends ExtensionHandler {
...
 
-      talismanAnalytics.capture("account create", { type: pair.type, method: "json" })
+      this.captureAccountCreateEvent(pair.type, "json")
     }
 
     return addresses
   }
 
-  private accountsCreateLedgerEthereum({
+  private async accountsCreateLedgerEthereum({
```
0xKheops: not sure if this one should be async, there is no await in there

apps/extension/src/core/libs/Analytics.ts
```diff
@@ -180,6 +175,13 @@ class TalismanAnalytics {
       .slice(0, 5)
 
     posthog.capture("balances top tokens", topChainTokens)
+    posthog.capture("balances report", {
+      $set: {
+        accountsCount: keyring.getAccounts().length,
```
0xKheops: this doesn't exclude watched accounts, not sure if intended or not

apps/extension/src/core/util/hasErc721Nft.ts
```diff
@@ -0,0 +1,38 @@
...
+}): Promise<Record<Address, boolean>> => {
+  const evmAddresses = keyring
+    .getPairs()
+    .filter(({ type, meta }) => type === "ethereum" && meta.origin !== AccountType.Watched)
+    .map(({ address }) => address)
+
+  const client = await chainConnectorEvm.getPublicClientForEvmNetwork(evmNetworkId)
+  if (!client) throw new Error(`Unable to connect to EVM network: ${evmNetworkId}`)
+
+  const data = await client.multicall({
```
0xKheops: multicall is automatic for all contract calls if network supports it, so here i think we should use a Promise.all([contract calls]) : 
- this method is network agnostic, and not all network support multicall.
- it would preserve typings for the results

0xKheops: ex here : https://github.com/TalismanSociety/talisman/blob/44a1066ee5e907844650ac39919c3fcf5f5617f9/apps/extension/src/core/util/decodeEvmTransaction.ts#L90

0xKheops (commented):
seems good, only minor comments

0xKheops (approved):
LGTM!

PR #1165, opened 2023-11-30
Support erc20 contracts with symbol method returning bytes32 type

chidg:


0xKheops (approved):
LGTM!

I'll do the same here
https://github.com/TalismanSociety/chaindata/blob/main/scripts/fetch-external/steps/fetchErc20TokenSymbols.ts

PR #1166, opened 2023-12-01
feat: new features highlight & wallet landing

alecdwm:


apps/extension/src/ui/domains/Asset/Buy/BuyTokensForm.tsx
```diff
@@ -140,6 +140,27 @@ export const BuyTokensForm = () => {
     address
   )
 
+  const autoSelectFirstAccountInit = useRef(false)
```
0xKheops, line 127: with metadata/opt-in PR this useAccounts will now suspense until all accounts are loaded, using a ref should become unnecessary then.
might want to add a TODO comment to remove it in the future

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioAccounts.tsx
```diff
@@ -364,18 +312,22 @@ export const PortfolioAccounts = () => {
     [balancesByAddress, currency, folder]
   )
 
+  const hasFunds = useMemo(() => balances.sum.planck.total > 0n, [balances])
+  const showGetStartedPopup = !hasFunds && ownedAccounts.length <= 2
```
0xKheops: balances aren't suspensed so this will flicker, sorry i said the contrary during the meeting
might want to use accounts (which will be suspensed with metadata PR) + the hasFunds flag that we have in appState (also suspensed)

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioAccounts.tsx
```diff
@@ -364,18 +312,22 @@ export const PortfolioAccounts = () => {
...
   return (
-    <FadeIn>
+    <FadeIn className="flex flex-col gap-12 pb-12">
       <Accounts
         folder={folder}
         folderTotal={folderTotal}
         portfolioOptions={portfolioOptions}
         watchedOptions={watchedOptions}
       />
+      {showGetStartedPopup && <NoAccountsPopup hasSomeAccounts={!!accounts.length} />}
```
0xKheops: might want to exclude watched accounts here, otherwise the thing disappears if we add one from "try it"

0xKheops (commented):
looks great, just few minor comments

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioWhatsNew.tsx
```diff
@@ -0,0 +1,111 @@
...
+          className="pointer-events-none relative w-full rounded-sm"
+          src={HeroUrl}
+          alt="a hero banner"
+        />
+        <div className="pointer-events-none absolute left-0 top-0 flex h-full w-full flex-col gap-4 p-8">
+          <div className="font-whyteInkTrap flex items-center gap-1 text-sm tracking-tight text-white">
+            <HandMonoLogo className="text-base" />
+            Talisman
+          </div>
+          <div className="text-primary text-2xl font-extrabold">V{process.env.VERSION}</div>
```
chidg: I don't think this should always show the latest version, but the version associated with the features listed. For example, if we release some major features in 1.X.0, it would be a bit confusing if this said 1.X.1, suggesting that the new features came out in the minor release.

To deal with this we could either code a mapping between "what's new message" and release version, or just go back to the idea of including the version number as part of the image rather than a separate html overlay.

apps/extension/src/ui/apps/popup/pages/Portfolio/PortfolioWhatsNew.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * This number should be incremented when we want to show the `What's New` banner
+ * to users who have dismissed it in the past.
+ *
+ * E.g. when we have new features to tell them about.
+ */
+export const WhatsNewVersion = 1
+
+export const PortfolioWhatsNew = () => {
```
chidg, line 30: It would be great if we could make this component generic, and just have it accept a JSON of values configuring the What's New message according to some standard format. This way anyone (@0byt for example) could submit a PR with a JSON object for a new message.

chidg (changes requested):

chidg (changes requested):
Some changes to the What's New setup requested

chidg (approved):
LGTM!

PR #1169, opened 2023-12-01
Asset display in popup fixes

chidg:
Fixes a few issues:
- selected account was not being properly selected in the popup, due to having two ways of managing selected account state. This PR consolidates both methods into one; in the popup, the selected account is now always determined by url parameters, while in the full screen, it's stored in a setting. This fixes the issue caused by this problem where all default tokens (DOT, KSM, ETH) would be displayed in empty accounts.
- when `hideDust` is true, this PR prevents the default tokens from being hidden in an empty account


alecdwm (approved):
LGTM!

PR #1171, opened 2023-12-04
fix: maintain `TokenLogo` and `ChainLogo` aspect ratio

alecdwm:
<div align="center">

image
**before**

image
**after**

</div>


0xKheops (approved):
LGTM!

PR #1172, opened 2023-12-04
Catch error in useAccountByAddress when invalid address given

chidg:
Caused an error when `?account=all` in URL params


alecdwm (approved):
Nice

PR #1174, opened 2023-12-05
Prevent rejection of ethereum auth request from dapp if there is a pending substrate one

chidg:


alecdwm (approved):
<div align="center">

lgtm

</div>

PR #1175, opened 2023-12-05
Allow blockeplorer to be null in wallet_addEthereumChain

chidg:
Follows viem's `AddEthereumChainParameter` more closely, and allows use with  https://metamask.github.io/test-dapp


0xKheops (approved):
LGTM!

PR #1182, opened 2023-12-08
build: use type import for types only library

tien:
This was causing build failure on the latest version of Vite


alecdwm (approved):
lgtm!

PR #1185, opened 2023-12-12
More staking reminders

chidg:


apps/extension/src/ui/domains/Portfolio/AssetsTable/AssetRowStakingReminderWrapper.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { ExternalLinkIcon, XIcon, ZapIcon } from "@talismn/icons"
+import { classNames } from "@talismn/util"
+import { useAnalytics } from "@ui/hooks/useAnalytics"
+import { useCallback } from "react"
+import { Trans, useTranslation } from "react-i18next"
+
+import { useStakingBanner } from "../../Staking/context"
+import { useTokenBalancesSummary } from "../useTokenBalancesSummary"
+
+export const AssetRowStakingReminderWrapper = ({
```
0xKheops: Finding this wrapper/render pattern pretty confusing.
wdyt about moving the code from line 23 to 32 into a new hook that would be called in DashboardAssetRow, and conditionally render the staking reminder from there ? this would remove the need for the staking reminder to wrap the row

0xKheops (commented):
LGTM 
- not approving as automerge is enabled, not sure if this should go in today's hotfix
- minor comment about component wrap pattern

alecdwm (commented):
LGTM2!

0xKheops (approved):
LGTM!

PR #1186, opened 2023-12-12
Prevent bug in analytics

chidg:


apps/extension/src/core/libs/Analytics.ts
```diff
@@ -174,14 +174,15 @@ class TalismanAnalytics {
       .map(({ chainId, tokenId }) => ({ chainId, tokenId }))
       .slice(0, 5)
```
alecdwm, line 175: A total nitpick -
It took me a hot second to figure out where the fiatSums are actually being sorted into highest-to-lowest balance order.
Maybe the sort on line `170` could be moved to after the filter on line `173`?

Doesn't really matter though

alecdwm (approved):
LGTM!!!

PR #1191, opened 2023-12-14
feat: assets discovery

0xKheops:


apps/extension/src/core/domains/ethereum/migrations/index.ts
```diff
@@ -0,0 +1,16 @@
+import { Migration, MigrationFunction } from "@core/libs/migrations/types"
+
+import { activeEvmNetworksStore } from "../store.activeEvmNetworks"
+
+export const migrateToNewDefaultEvmNetworks: Migration = {
+  forward: new MigrationFunction(async () => {
+    await activeEvmNetworksStore.set({
```
chidg, line 7: I don't think a migration is the best way to do this, since the store is only being introduced in this PR. The migration would only be applied for people upgrading from an old version, which means we'd need to set this initial data for people installing newly from this release onwards in a different way. If the store is always meant to be created with a particular set of data, it would be better to handle that somewhere else, eg: the constructor function of the store, rather than in a migration.

chidg (changes requested):
Absolutely *fantastique*
One comment re: migration

chidg (commented):
Absolutely *fantastique*
One comment re: migration

apps/extension/src/core/domains/ethereum/migrations/index.ts
0xKheops: Made it so that new installs (fake migrations) get an empty store, but existing installs (real migration) would get store initialize with data that reenables old defaut networks that arent default anymore.
Did i get this wrong?

chidg (approved):
LGTM

PR #1192, opened 2023-12-14
fix: layout shift on dashboard asset rows

0xKheops:
- Fixes a layout shift on dashboard asset rows that was introduced by #1171


chidg (approved):
LGTM

PR #1199, opened 2023-12-19
fix: incorrect instantiation of `ChainConnectorEvm`

tien:


0xKheops (dismissed):
LGTM!

chidg:
A little hard to tell because of the way `ChainConnectorEvm` can be instantiated in an arbitrary balance module, but is it possible that we could change the constructor signature to only take a single argument for `chaindataEvmNetworkProvider` and `chaindataTokenProvider` rather than two? Since it appears that the same variable is supplied for both args in the cases where it's used. @alecdwm

chidg (approved):
LGTM

PR #1202, opened 2023-12-21
feat: sign in with substrate

alecdwm:
- [x] Show SIWS UI when signing SIWS messages with the wallet
- [x] Fix ledger raw message signing so that we can support SIWS for Polkadot ledger accounts


chidg (approved):
LGTM!

PR #1203, opened 2023-12-22
fix: invoke stopListening when an error is caught

chrisling-dev:
Hey @0xKheops  seems like a bug here, the `stopListening` call wasn't invoked when an error is caught. feel free to close this if this was intended


0xKheops (approved):
LGTM! thanks

PR #1204, opened 2023-12-27
Feat/import from signet

chrisling-dev:


apps/extension/src/core/domains/accounts/types.ts
```diff
@@ -62,13 +62,20 @@ type AccountJsonDcentOwnProperties = {
 
 export type AccountJsonDcent = AccountJson & AccountJsonDcentOwnProperties
 
+type AccountJsonSignetOwnProperties = {
+  signetUrl: string
```
chrisling-dev, line 66: Since Signet can be self hosted, we want to make sure it's easy to identify which url a vault was imported from so users with many vaults can easily locate their vaults

chrisling-dev (commented):
Self reviewed

apps/extension/src/core/domains/signing/handler.ts
```diff
@@ -171,6 +174,33 @@ export default class SigningHandler extends ExtensionHandler {
...
+      genesisHash: queued.account.genesisHash || "",
+      dapp: url,
+    })
+
+    // close popup so Signet signing page can be open in full screen normal browser
+    // users will most likely stay on Signet anyway to review the pending tx
+    // so the popup is not needed here and can be closed
+    windowManager.popupClose()
+    await Browser.tabs.create({
+      url: `${queued.account.signetUrl}sign?${params.toString()}`,
```
chidg: I noticed that there doesn't seem to be any validation on the input for the Signet url initially. That means there could be an issue here, if this url construction expects that the `signetUrl` will end with a `/` but it doesn't. Perhaps we should construct a url using `new Url()` here, to ensure it's formatted correctly, as well as doing a bit of basic regex validation in `ConnectSignetPage`

apps/extension/src/ui/domains/Account/AccountAdd/Container.tsx
```diff
@@ -185,6 +185,12 @@ const ConnectAccountMethodButtons = () => {
         networks={["polkadot", "ethereum"]}
         to={`/accounts/add/dcent`}
       />
+      <AccountCreateMethodButton
```
chidg, line 188: Currently this component `AccountAdd/Container` is used directly in both the onboarding and the dashboard. So, the Signet option shows up in both places, however because this PR only implements the `AccountAddSignet` components in the dashboard, the onboarding version is just a dead link. You can decide whether you want the option to connect to Signet to be present in the onboarding, and implement the next screens, or else modifiy this component to detect if it's in the onboarding state and hide the signet option.

chidg (commented):
Great stuff Chris, couple of improvements requested.
I made a small commit too (it was part of a bigger one, but I realised that was a bad idea!)

apps/extension/src/core/domains/signing/handler.ts
chrisling-dev: Hey thanks for reviewing! Actually there is validation on input!

1. The button is disabled if `signetUrlOrigin === ""`:
https://github.com/TalismanSociety/talisman/pull/1204/files#diff-3c5cd66280dad759467aba78e5178233e7f7d866af201c04ec6b154a31dd87ebR109

2. `signetUrlOrigin` will only have a value if it is a valid url
https://github.com/TalismanSociety/talisman/pull/1204/files#diff-9ec58f60f5ee94c75c9dc9e489f3803c3a60532d71370e3f11108f895e2caec4R13-R21

Note that `signetUrlOrigin` is returning `url.href`, which will always have an ending `/`

```javascript
console.log(new URL("https://signet.talisman.xyz/").href)
// 'https://signet.talisman.xyz/'

console.log(new URL("https://signet.talisman.xyz").href)
// 'https://signet.talisman.xyz/'
```

We could still construct a new URL here and add validation but I reckon that might be redundant considering `signetUrlOrigin` is the only entry point to create a signet account, lmk if this validation on `ConnectSignetPage` is sufficient or if you still think we should have stronger validation and I'll be happy to add it

Another reason I did not add very strict validation is because clients could be self hosting on existing domains under different paths (e.g. https://clientdomain.com/signet/...signet-paths), hence I simply extracted whatever parts there is before "overview", as that is likely the default path users will copy signet from which is the dashboard. (https://signet.talisman.xyz/overview)

Just UX wise tho I think most users will not even touch the signet url manually at all because it is prefilled, so it's really only for self hosting clients that most likely know what they're doing / will reach out to us for help if there is an issue

apps/extension/src/ui/domains/Account/AccountAdd/Container.tsx
chrisling-dev: good catch, wasn't aware it's in the onboarding flow, will look 👀

chrisling-dev: Okay added Signet connect to the onboarding flow and tested!

apps/extension/src/core/domains/signing/handler.ts
chrisling-dev: Hey @chidg so instead of the UI solution, I just make it 
- use origin only and removed the logic to extract parts before overview - I realized even if we handle self hosting on sub-path, most operations in signet wont be supported anyway (e.g. we always redirect to /overview instead of /your-path/overview), so it makes no sense to support sub path.
- Added function to add trailing slash dynamically so it doesnt matter whether we're using `href` or `origin`

chidg (dismissed):
LGTM!

chidg (approved):
LGTM!

PR #1205, opened 2023-12-28
fix: connected includes removed addreses

chrisling-dev:
This PR fixes a bug where the `connected` count is including removed addresses:

Screenshot 2023-12-28 at 7 55 33 PM

The proper fix should be to remove address from `sites[id].addresses` (or `ethAddresses`) when an address is removed.


chidg (approved):
LGTM!

PR #1210, opened 2024-01-03
fix: faster QR scanning

0xKheops:
Fixes a problem that i found while investigating on issue 1198 (which i could not reproduce)

the white padding around the QR wasn't big enough, resulting in my android PV scanning only 1 frame every few seconds
with a bigger padding it now scans very fast, at least 5+ frames per sec


alecdwm (approved):
LGTM ✨

PR #1212, opened 2024-01-03
fix: improve formatting for fiat values between `$0.00` and `$0.01`

### LINKED ISSUE #1148: Add decimal points to assets with low prices
**Describe the bug**
Some assets have very low prices, and we cutoff the price at 2 decimal places, making the price indicator unusable for users. 
For prices that have many trailing zeros, can we add an extra decimal place or two?

**Links**
[If applicable, add links to Sentry issues, Upvotey bug reports, or Discord messages.](https://discord.com/channels/858891448271634473/879997657114882078/1171929338535485600)

**Screenshots**
If applicable, add screenshots/recordings to help explain your problem.
image

image

alecdwm:
Closes https://github.com/TalismanSociety/talisman/issues/1148

For fiat values like `0.0005618` we currently render `$0.01`.
With this PR we render `$0.00056` instead.


0xKheops:
It's a great upgrade for token price, and gas when signing a tx, but i think for balances it should keep displaying `< 0.01$`
wdyt ? 

ex below where it feels weird
image

alecdwm:
Yeah I agree, `< $0.01` is just fine for balances!

apps/extension/src/ui/domains/Asset/Fiat.tsx
```diff
@@ -74,11 +52,62 @@ export const Fiat = ({
...
+  currencyDisplay,
+  isBalance,
+  noCountUp,
+}: DisplayValueProps) => {
+  /**
+   * Represents the first non-zero decimal place for numbers between -1 and 1
+   * For example:
+   *
+   *     Input  - Output
+   *     ****** - ******
```
chidg: This is a great comment, but we could improve it by making clear what the actual output is - eg `99` instead of `0` for input `99.000`

alecdwm: Hmmmmm, the output of `99.000` should be `0`. Are you certain that you're getting `99` from it?

alecdwm:
```shell
> Math.abs(Math.min(0, Math.floor(Math.log10(Math.abs(99.000)))))
0
```

chidg (approved):
LGTM!

PR #1213, opened 2024-01-04
perf: better qr generation performance

0xKheops:
Enhances performance of the QR code generation, from 28s to 26s for a 33K frames test payload.

Basically It reduces the amount of re-renders & reduces the computation done on each render iteration, allowing more resources to be used on the QR generation


alecdwm (approved):
LGTM ser

PR #1217, opened 2024-01-04
fix: substrate transactions stuck in pending state

### LINKED ISSUE #1195: wallet tx history : substrate tx stuck in pending state
timeout of 90s doesn't seem to be working

0xKheops:
- Fixes #1195

Substrate transactions that are still `pending` after subscriptions closed will have their status set to `unknown`


chidg (approved):
LGTM

PR #1222, opened 2024-01-10
feat: cached balance totals

0xKheops:
Increases performance of the popup home page by caching balance totals for each account, which removes the need to load tokens/networks/balances to display that page.

Notes: 
- Converted most of our context providers into recoil atoms as they did leverage tokens/networks/balances and were triggering their associated suspense


chidg (dismissed):
LGTM!

alecdwm (approved):
<div align="center">

LFG
**LFGooooooooooooooooooo**

</div>

PR #1226, opened 2024-01-10
feat: rework asset discovery page

0xKheops:
- replaces buttons by toggles
- copy update
- better table layout (last column has fixed width)


chidg (approved):
LGTM!

PR #1237, opened 2024-01-12
feat: asset discovery page polishing

0xKheops:
- Virtualized rows rendering
- Sorted rows by network then symbol
- Token rates are now fetched by batch of 100 and stop fetching in case of 429


chidg (approved):
LGTM

PR #1241, opened 2024-01-15
fix: popup header flickering on authorised sites

0xKheops:
Fixes the flickering caused by the account & network selector, when opening embedded popup from an authorised tab

This PR introduces the useRecoilPreload utility which allows to preload all the atoms used by a given screen, for a faster first render. 

I also added a log.debug on each atom initialization, do help developer troubleshot suspense issues, and/or identify which atoms should be preloaded by a screen.


chidg (approved):
LGTM

PR #1243, opened 2024-01-16
feat: redirect to existing network form

0xKheops:
also includes copy change for the toggle from "Display balance" to "Active" (+ tooltip)


alecdwm (approved):
🔥🔥🔥

PR #1248, opened 2024-01-17
feat: suspense rampage

0xKheops:
checked for screens that trigger suspenses all around the app to remove "flashing screen" behaviors

- added a bunch of useRecoilPreload to load everything needed at once (send funds now opens fast!)
- added missing suspenses (QrCode...)
- changed accounts screen to leverage balanceTotals instead of balances
- converted useAccountByAddress and useChainByGenesisHash to recoil family selectors


chidg (approved):
LGTM

PR #1249, opened 2024-01-18
chore: update whats new text

alecdwm:
Now includes extra 🌶️ spice 🌶️🌶️🌶️🥵 in the form of  
✨links✨ 🔗🖱️🌐 and  
✨buttons that look like links✨ 🧙🤔🔍

<div align="center">

full
**the full description**

scrollable
**what users will see before scrolling**

</div>


chidg:
Looks great but - I think the `Check it out` link would look better justified right?
and let's put something in about perf:

**🔥Blazing Fast Performance:** We've done a major overhaul of wallet performance. Talisman is now faster and smoother than ever before.

chidg (approved):
LGTM

PR #1252, opened 2024-01-18
feat: support for acala evm+ encoded gas settings

0xKheops:
Waiting for acala team to explain how to compute fee (to display to the user, and prevent submitting tx if not enough funds)
I suspect this is not possible so submitting this PR For review now, and will do another one if necessary


chidg (approved):
LGTM!

PR #1253, opened 2024-01-18
fix: add custom token popup for known tokens

0xKheops:
- allows the add token popup to show up for known tokens that are currently disabled
- enforces active state when adding custom networks, to prevent them from disappearing if reset to default chaindata state
- clear its active state after removing a custom token


chidg (approved):
LGTM

PR #1255, opened 2024-01-18
fix: prevent double dapp auth prompt

0xKheops:
creates the request object synchronously, allowing the tabs handler to reject duplicate auth requests from dapps


chidg (approved):
LGTM

PR #1258, opened 2024-01-21
feat: azero.id

alecdwm:
I think we should feature crit this before merge to check that everyone is cool with:

- Aleph logo on orbs for azero.id accounts

<p align="middle">
exhibit a</img>
exhibit b</img>
</p>

- Copy used on the send funds warning when the account format / azero id doesn't match the chain used for the transfer

<p align="middle">
exhibit c</img>
exhibit d</img>
</p>


0xKheops (commented):
LGTM (pending yarn.lock fix)

0xKheops (approved):
LGTM

PR #1260, opened 2024-01-23
chore: small `ChaindataProvider` refactor

alecdwm:


packages/chaindata-provider/src/ChaindataProvider.ts
```diff
@@ -1,37 +1,700 @@
...
+  }
+
+  //
+  // items by genesisHash
+  //
+
+  get chainsByGenesisHashObservable() {
+    return this.chainsObservable.pipe(map(util.itemsToMapByGenesisHash))
+  }
+  get chainsByGenesisHash() {
```
0xKheops: not super comfortable with async properties, callers may think they can call it multiple times without impact. 
would prefer a getChainsByGenesisHash() method.
wdyt ?

alecdwm: I think that's a good idea, I'll change them all to be getXXX methods

alecdwm: Done

0xKheops (dismissed):
great small refactor :)

chidg (approved):
LGTM

PR #1261, opened 2024-01-23
fix: evm network reset

0xKheops:
couldn't come up with a safe way to update the form's value upon reset, so i've put back the logic to navigate to the networks list screen


chidg (approved):
LGTM

PR #1265, opened 2024-01-23
fix: redirect to existing network

0xKheops:
existing network modal will now use a navigation history replace when redirecting to the existing network page, preventing the user to run into the same existing network modal if he clicks the back button

also took the opportunity to highlight the network name, like we do for other similar modals


chidg (approved):
LGTM

PR #1266, opened 2024-01-23
feat: token type pill on the token picker

0xKheops:
added a token type pill in the token picker as some ERC20 & native tokens have same symbol 

note: the token picker is used by send funds, copy address modal (for now), and buy tokens modal
note 2: the pill only displays for EVM Native & ERC20 tokens, substrate ones are ignored

image


chidg (approved):
LGTM

PR #1272, opened 2024-01-25
chore: change native token ids

0xKheops:
removes the symbol from native token ids

TODO : 
- changes in chaindata
- regenerate init files


apps/extension/src/ui/atoms/chaindata.ts
```diff
@@ -322,7 +329,31 @@ export const allTokensMapState = atom<TokenList>({
   effects: [
     ({ setSelf }) => {
       log.debug("allTokensMapState.init")
-      const sub = chaindataProvider.tokensListObservable.subscribe(setSelf)
+      const sub = chaindataProvider.tokensListObservable.subscribe((tokensList) => {
+        // ignore native tokens duplicates, this will also ignore duplicate balances
+        // this is necessary for users that had created custom native tokens prior we changed native token ids (v1.21.0)
+        // note: this cleaup takes 3ms on my machine
```
chidg: I think regardless of performance impact, this would be better off in a migration, since it should only have to happen once, and then the code can live in a nice easily maintainable place

chidg (commented):
LGTM, although I think once-off operations should go into migrations, so that business logic can be simplified

chidg (approved):

chidg (approved):
LGTM

PR #1281, opened 2024-01-26
fix: onboarding & account import fixes bundle

0xKheops:
- triggers chaindata hydratration as soon as onboarding opens, so account import pickers can load balances right away
- any balance subscription will wait for chaindata to be hydrated before initializing the subscription
- account import pickers now fetch all balances from all active tokens (not restricted to dot/ksm/glmr/astar anymore)
- account import pickers balances get stale after 30 sec if still initializing (same as wallet balances)


chidg (approved):
LGTM

PR #1283, opened 2024-01-29
Prevent minimetadata updating for custom chains if no substrate accounts

chidg:
Thanks to @alecdwm for doing all the hard work ;)


alecdwm (approved):
Lgtm 😆

PR #1284, opened 2024-01-30
fix: extension tests issues

alecdwm:
problems fixed:
- pjs keyring using a different prefix in tests to dev/prod
- tests not waiting for `@talismn/scale` to be ready before they execute
- chaindata-provider-extension not mocked after running `yarn build:packages`
- `awaitKeyringLoaded` sometimes never resolving after keyring did in fact load
- calls to `awaitKeyringLoaded` never cleaning up their subscription to the keyring.accounts subject


apps/extension/webpack/webpack.common.js
```diff
@@ -115,8 +115,6 @@ const config = (env) => ({
   plugins: [
     new webpack.DefinePlugin({
       // passthroughs from the environment
-      "process.env.EXTENSION_PREFIX": JSON.stringify(""), // this env var MUST be set, however if it has a value the keyring will break
-      "process.env.PORT_PREFIX": JSON.stringify(process.env.PORT_PREFIX || "talisman"),
```
alecdwm: ~~These never did anything. The `DefinePlugin` will replace the literal code `process.env.EXTENSION_PREFIX` with `""`, but this doesn't exist anywhere!~~

~~Instead, pjs deps do a process.env["PORT_PREFIX"]~~

I stand corrected 😆 
The pjs source code does `process.env["PORT_PREFIX"]`, but the distributed code does `process.env.PORT_PREFIX`.

chidg (approved):
LGTM

PR #1296, opened 2024-02-02
fix: throttle balance totals

0xKheops:
balance totals are now computed using throttle instead of debounce, allowing for much quicker updated totals after adding accounts


chidg (approved):
Much better, LGTM

PR #1297, opened 2024-02-05
feat: coingecko api key

0xKheops:
- removes FeatureFlagsStore
- adds new RemoteConfigStore
- uses dedicated config (url, header name, api key) pulled from the config repo on each coingecko request


chidg (approved):
LGTM!

PR #1298, opened 2024-02-07
fix: tokenRates migration

alecdwm:
Previous migration kills the crab in prod

New migration will only run for users who either (a) haven't upgraded yet, or (b) encountered the db error after upgrading


chidg (approved):
LGTM

PR #1300, opened 2024-02-07
fix: allow signing eip712 message with invalid verifying contract

0xKheops:
We introduced a validity check on the verifying contract of an EIP712 during latest audit.
The audit feedback was about possibility to execute arbitrary javascript when copying it's address, though we thought it would be best to check that the address is a valid EVM address, unlike metamask which doesn't.

It turns out that veryfing contract address may not always be a evm address.
Our valid address check breaks features on some dapps such as https://portal.dymension.xyz/ 

This PR removes the evm valid address check, in favor of a .startsWith("javascript:") check

### DIFF
--- a/apps/extension/src/ui/domains/Sign/SignRequestContext/EthereumSignMessageRequestContext.ts
+++ b/apps/extension/src/ui/domains/Sign/SignRequestContext/EthereumSignMessageRequestContext.ts
@@ -2,7 +2,6 @@ import { KnownSigningRequestIdOnly } from "@core/domains/signing/types"
 import { log } from "@core/log"
 import { HexString } from "@polkadot/util/types"
 import { provideContext } from "@talisman/util/provideContext"
-import { isEthereumAddress } from "@talismn/util"
 import { api } from "@ui/api"
 import { useEvmNetwork } from "@ui/hooks/useEvmNetwork"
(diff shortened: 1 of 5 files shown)


apps/extension/src/ui/domains/Sign/SignRequestContext/EthereumSignMessageRequestContext.ts
```diff
@@ -44,7 +43,8 @@ const useEthSignMessageRequestProvider = ({ id }: KnownSigningRequestIdOnly<"eth
       // for now only check signTypedData's verifying contract's address
       const typedMessage = isTypedData ? JSON.parse(request.request) : undefined
       const verifyingContract = typedMessage?.domain?.verifyingContract as string | undefined
-      if (verifyingContract && !isEthereumAddress(verifyingContract)) return false
+      if (verifyingContract && verifyingContract.toLowerCase().startsWith("javascript:"))
```
chidg, line 46: AFAICT all the other instances of the verifyingContract derive from this hook - so, it should be safe to validate here only (or else, upstream of here eg. in the tabs handler) and that would be more easily mantainable

chidg (approved):
LGTM, validation to be refactored in future

PR #1303, opened 2024-02-08
fix: migration errors bundle

0xKheops:
- clears the database if there is an error occuring while changing id of balance entries as part of the db migration
- ensures the db error dialog can display


chidg (approved):
LGTM

PR #1311, opened 2024-02-13
feat: upgrade `@talismn/balances-react` to github chaindata

alecdwm:


packages/balances/src/util/hydrateChaindata.ts
```diff
@@ -0,0 +1,47 @@
...
+export const updateCustomMiniMetadata = async (
+  chaindataProvider: ChaindataProvider,
+  miniMetadataUpdater: MiniMetadataUpdater,
+  isTest?: boolean
+) => {
+  // Don't update custom minimetadata in tests
+  //
+  // TODO: Remove this, and instead mock the websocket response for all of the called rpc methods.
+  // E.g. state_getMetadata, system_properties, etc
+  if (isTest) return
```
chidg: Checking for test state in the code itself is definitely not a good way to do things, so I would really prefer not to elevate that into a function argument. If we can't mock it, can we stick with the existing constant check inside the function?

alecdwm: The problem with that is that the `TEST` constant comes from the extension package, which depends on the `@talismn/balances` package where this code is - not the other way around.

chidg (approved):
LGTM!

PR #1316, opened 2024-02-20
feat: shorten copy flow

### LINKED ISSUE #1229: Yeet the receive flow
Replace receive flow by the copy address flow

0xKheops:
- Closes #1229 

Changes :
- Removes the receive flow, only the copy address flow is remaining
- Direct copy for ethereum and chain specific substrate accounts, to skip the wizard if unnecessary
- Displays address and type of accounts in the copy wizard account picker
- Show QR code button in address copied notifications
- "Receive" labels changed to "Copy"

Misc : 
- removes the unnecessary space on the right for ScrollContainers (ex: popup home with more than 5 accounts, token pickers, etc)


chidg (approved):
LGTM

PR #1319, opened 2024-02-22
Misc evm UI improvements

chidg:


apps/extension/src/ui/apps/dashboard/routes/Settings/Mnemonics/MnemonicsPage.tsx
```diff
@@ -191,9 +193,14 @@ const MnemonicRow: FC<{ mnemonic: Mnemonic }> = ({ mnemonic }) => {
                   )}
                 </div>
               </ContextMenuItem>
-              <ContextMenuItem onClick={handleSetVerifierClick} disabled={isVerifier(mnemonic.id)}>
-                {t("Set as Polkadot Vault Verifier Certificate")}
-              </ContextMenuItem>
+              {hasQrAccount && (
```
0xKheops: can we change the test to hasVerifierCertificateMnemonic instead of hasQrAccount ?

if user already has a verifier mnemonic (maybe accidentally clicked yes to generate a new one from the "metadata portal" page), and wants to the one he's been using on his PV afterwards, there is now way to do it before adding an account. 

Since verifier mnemonic is created as soon as user adds a PV account, it should be the best of both worlds

0xKheops (commented):
LGTM aside from the minor comment below

0xKheops (approved):
LGTM!

PR #1320, opened 2024-02-23
fix: prevent login prompt from opening twice

0xKheops:
to reproduce the bug : 
- run Talisman in dev mode (so strict mode will trigger the login prompt twice)
- bookmark the dashboard
- close browser
- open dashboard from the bookmark


chidg (changes requested):
Not sure about this one. The Window Manager should be the place where we keep state about which windows are open and logic about when or if to open them, rather than the app handler. Would that be possible?

chidg (approved):
LGTM

chidg (approved):
LGTM

PR #1324, opened 2024-02-27
feat: settings routes structure

0xKheops:
Modified routes structure so that child screens are leafs of their parent screens.
This allows for the current section's link in sidebar to remain highlighted when browsing a level 2 settings page.

Note : added redirects from old urls to their associated new urls just in case i forgot to rewrite a link along the way


chidg (approved):
LGTM

PR #1326, opened 2024-02-28
fix: send funds prevent address type mismatch

0xKheops:
fixed the recipient picker validation so a pasted EVM address can't be selected when transfering substrate assets


chidg (approved):
LGTM!

PR #1327, opened 2024-02-28
fix: less portfolio suspenses

0xKheops:
- removes suspenses when navigating in portfolio
- removes suspenses when changing account or network in portfolio
- network filter logic fixed


chidg (approved):
Very snappy. LGTM

PR #1331, opened 2024-03-01
Minor send funds components refactor

chidg:
- Moved drawer components to common location, and draw-control behaviour to context hook (in preparation for new drawer from #1258
- Extracted sub-components of `SendFundsAmountForm` to live in separate files


apps/extension/src/ui/apps/popup/pages/SendFunds/context.ts
```diff
@@ -22,6 +49,7 @@ export type SendFundsWizardPage = "from" | "to" | "token" | "amount" | "confirm"
 const useSendFundsWizardProvider = () => {
   const [searchParams, setSearchParams] = useSearchParams()
   const navigate = useNavigate()
+  const drawers = useSendFundsDrawers()
```
chidg: I put this into the Wizard context rather than the `useSendFunds` context hook because I figured it was more related to UI state than the state of the send funds transaction itself

0xKheops: this context is meant to store data must be carried from one screen of the wizard to another, and navigation logic.
my understanding is that both of these drawers are local to their parent component/screen, and don't need to be exposed to other components of the wizard, so i don't think that this `drawers` property belongs in any of the 2 contexts.

for example for the add contact scenario, useOpenClose could be called from AddContext.tsx, and isOpen+close would be simply passed as prop to the child drawer
same for ReviewButton=>ForfeitDrawer (but there the close callback would to be wrapped in a method that also calls goToReview afterwards)

wdyt ?

0xKheops (approved):
LGTM!

PR #1343, opened 2024-03-06
refactor: background package

0xKheops:
Extracts the backend code in it's own `extension-core` package, enabling the following use-cases : 

- export only the code that is intented to be reused by UI
- domain-specific ui packages. 

Also added a `extension-shared` package containing constants & log helper. This allows sharing this part between extension, extension-core, and most importantly with `page.js` & `content_script.js`.
`extension-shared` supports treeshaking, unlike `extension-core`.


chidg (approved):
LGTM

PR #1346, opened 2024-03-07
refactor: remove @polkadot/util-crypto peer deps on @talismn/orb

0xKheops:
Removed peer dependency on `@polkadot/util-crypto` to make it easier to integrate on dapps.

Main reason is `@polkadot/*` packages have version requirements on each other, making it painful to tinker with.
Other reason is to remove the need for EVM dapps to add polkadot-related dependencies.


0xKheops:
@alecdwm do you think this would crash on browsers with limited wasm support ?

I don't remember which bit caused that few months ago, nor how to test it.

0xKheops:
nevermind @alecdwm 
removed wasm stuff

chidg (approved):
Much simpler
LGTM

PR #1348, opened 2024-03-08
feat: balance summary tooltips

0xKheops:
Improves tooltips of balance details on account import UIs that display a fiat total per account (mnemonic, ledger, json & qr imports)

Before: 
image

After:
image


packages/util/src/isNotNull.ts
```diff
@@ -0,0 +1 @@
+export const isNotNull = <T>(x: T | null | undefined): x is T => !!x
```
chidg: It seems a bit confusingly named, since it really checks for boolean truthiness, not only 'not nullness'.
Alternative names:
`isBooleanTrue`
`trueTypeGuard`
`isReallyAThing`

chidg (commented):
Looks very nice, one (nitpicky) comment about naming of utility function

packages/util/src/isNotNull.ts
0xKheops: good point, there are few edge cases that i didn't take into account such as 0 (number) or empty string...
my intent was to get invalid types trimmed out when using it as an array filter callback, as .filter(Boolean) doesn't do anything :/

will go with isBooleanTrue

chidg (approved):
LGTM

PR #1349, opened 2024-03-08
fix: incorrect origin check with browser prerendering

### LINKED ISSUE #1279: Incorrect origin due to browser pre-rendering
**Describe the bug**
Pre-rendering of pages by browsers can cause incorrect origins to be passed to the extension, prompting for authorization from the site the page got prerender from or throwing an error in the case of urls like chrome://newTab. This can happen when hovering links or while typing in the browser URL bar. 

Please see discussion in https://github.com/polkadot-js/extension/issues/1293

As Talisman also supports iframes I think using the `sender.url` is also safer than using the `sender.tab.url`. That way if the iframe that interacts with the wallet is swapped out the user would get a new authorization request to review.

**To Reproduce**

1. Preloading must be enabled in chrome://settings/performance
2. open a random site or new tab
3. in the address bar start typing a page that uses the extension. 
4. If it's a site you frequent chrome will likely start to preload and pre-render it as you are typing the url. 
5. depending on how the site handles the error the extension it may throw prompt for authorization from the site your visited from or in the case of a new tab throw an error. See below video. 

**Video**

https://github.com/TalismanSociety/talisman/assets/42175565/c4cb06f7-0ad1-4cd3-956a-3109f94a7528

**Expected behaviour**
the page the user is visiting should be the one prompting for authorization. should not cause an error when visiting a dApp from a new Tab.

**Links**
https://github.com/polkadot-js/extension/issues/1293

**Environment (please complete the following information):**

- OS: iOS or Windows
- Browser: Chrome 
- Version: Version 121.0.6167.85

**Potential Solution**
Editing [this line](https://github.com/TalismanSociety/talisman/blob/460fa6f504c44b50b9600a580aae33a4324bbc40/apps/extension/src/core/handlers/index.ts#L53) to use the `sender?.url` instead of the `sender?.tab?.url` resolves the issue.

For engineering:
**Test**
Should a test be written to accompany the bug fix that squashes this bug? If not, what is the reasoning for not including a test case?

0xKheops:
- Fixes #1279


chidg (approved):
LGTM

PR #1353, opened 2024-03-11
feat: mobile friendly ux

0xKheops:
- enforces a min width on main content of all screens, and makes main area scrollable, to ensure wallet is usable even on very small resolutions
- modal dialog max size is the size of the screen 
- removes the 6-columns layout for 24 words mnemonics, which was causing some mnemonic words to be unreadable


chidg (approved):
LGTM

PR #1359, opened 2024-03-20
feat: block explorer & coingecko links :link: 

0xKheops:


apps/extension/src/ui/apps/dashboard/routes/Portfolio/AccountContextMenu.tsx
```diff
@@ -115,35 +132,38 @@ export const AccountContextMenu = forwardRef<HTMLElement, Props>(function Accoun
...
+        <Suspense>
+          {account && (
+            <>
+              {canToggleIsPortfolio && (
+                <ContextMenuItem onClick={toggleIsPortfolio}>{toggleLabel}</ContextMenuItem>
+              )}
+              {canCopyAddress && (
+                <ContextMenuItem onClick={copyAddress}>{t("Copy address")}</ContextMenuItem>
+              )}
+              {account && <ViewOnExplorerMenuItem account={account} />}
```
chidg: Don't need a check for `account` here, it's already inside one.
```suggestion
              <ViewOnExplorerMenuItem account={account} />
```

apps/extension/src/ui/domains/ViewOnExplorer/ExplorerNetworkPicker.tsx
```diff
@@ -0,0 +1,156 @@
...
+                type: "substrate",
+                name: chain.name!,
+                explorerUrl: chain.explorerUrl!,
+              })
+            )
+        : [],
+    [address, evmNetworks]
+  )
+
+  const sortedNetworks = useMemo(
```
chidg, line 75: Maybe not worth worrying about for this time, but perhaps in future we could extend the sorting method to include crypto balances with no fiat value - so, first up would be chains with fiat value, then next would be chains with some token balances without any known fiat value, then finally chains sorted in alphabetical order.

chidg (approved):
LGTM

PR #1360, opened 2024-03-21
fix: connect signet dead object

chrisling-dev:
Firefox blocks access to the opened tab. This PR handles that case gracefully by catching the error and toasting "Blocked by browser" message


alecdwm (approved):
LGTM!

PR #1362, opened 2024-03-22
feat: add pool ID to balance metadata

tien:


packages/balances/src/modules/SubstrateNativeModule.ts
```diff
@@ -1010,6 +1010,7 @@ export async function subscribeNompoolStaking(
       ]) => {
         const balances: SubNativeBalance[] = Array.from(poolIdByAddress).map(
           ([address, poolId]) => {
+            const parsedPoolId = poolId === null ? undefined : parseInt(poolId)
```
tien, line 1013: Why is this nullable, btw? There must always be a pool ID, right?

alecdwm: Comes from [here](https://github.com/TalismanSociety/talisman/pull/1362/files?w=1#diff-1cdeeb2aa917792c75edc6288720eb15b435868a210785fb0155bad72fe73a7eL808-L809), i.e.:

```ts
const poolId: string | undefined = decoded?.poolId?.toString?.()
const points: string | undefined = decoded?.points?.toString?.()
```

I don't think we should trust the RPC responses to (a) be the same for every chain, nor (b) not change in a future version.  
There have already been discrepancies in the shape of the data in a few of the other balances queries we make, so I don't think I'm being paranoid in that.  
And in the event `poolId` really does end up being undefined, I want our code to handle that instead of crashing the thread.

alecdwm (approved):
LGTM!

PR #1363, opened 2024-03-23
fix: evm tx with low base fee

0xKheops:
- fixes invalid custom fee settings default value when base fee is very low
- fixes custom gas fee form to support very low base fee
- fixes transaction validation on rollup with very low base fee (manta pacific, scroll, zora, etc.)

Misc : migrated codebase to use viem v2 and wagmi v2


chidg (approved):
LGTM

PR #1364, opened 2024-03-25
fix: frozen token amounts in `SubAssetsModule`

alecdwm:


chidg (approved):
Confirmed that DED appears, and that it is displayed as a locked balance and not transferrable.
LGTM!

PR #1369, opened 2024-03-25
fix: evm tx browser notification

0xKheops:
Fixes old regression from ethers => viem migration

Noticed the while performing a swap that reverted. Browser/OS notification stated that transaction confirmed, while it didn't.


chidg (approved):
LGTM

PR #1375, opened 2024-03-26
fix: copy modal - restrict to evm accounts for evm networks

0xKheops:
on asset details page while no account is selected (ex: portfolio => all accounts => ETH), when clicking the copy button for an EVM chain, modal was prompting user to select from all accounts including substrate ones

this PR filters out substrate accounts if clicking the copy address button for an EVM network


chidg (approved):
LGTM

PR #1376, opened 2024-03-26
fix: send funds recipient form - restrict onChainId options based on context

0xKheops:
prevents ENS resolution when sending substrate tokens


chidg (approved):
LGTM

PR #1379, opened 2024-03-28
fix: dev sourcemaps

alecdwm:


apps/extension/webpack/webpack.dev.js
```diff
@@ -15,7 +15,7 @@ const faviconsSrcPath = path.join(__dirname, "..", "public", "favicon*.*")
 
 const config = (env) =>
   merge(common(env), {
-    devtool: "inline-cheap-module-source-map",
+    devtool: "eval-cheap-module-source-map",
```
alecdwm: this is the fix

config/tsconfig/base.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "display": "Default",
   "compilerOptions": {
+    "sourceMap": true,
```
alecdwm, line 5: i added this because [it's supposed to be necessary](https://github.com/TypeStrong/ts-loader?tab=readme-ov-file#devtool--sourcemaps), although it didn't seem to have any effect on the dev mode sourceMaps. maybe `ts-loader` is implicitly setting it to true or something

apps/extension/webpack/webpack.dev.js
alecdwm: inline is only good for single-module builds or something

0xKheops (approved):
LGTM :heart:

PR #1380, opened 2024-03-28
Balances Dedexification

chidg:


packages/balances/src/types/balances.ts
```diff
@@ -255,6 +255,17 @@ export class Balances {
...
 
+export const getBalanceId = (
+  balance: Pick<
+    BalanceJson,
+    "source" | "subSource" | "address" | "chainId" | "evmNetworkId" | "tokenId"
+  >
+) => {
+  const { source, subSource, address, chainId, evmNetworkId, tokenId } = balance
+  const locationId = chainId !== undefined ? chainId : evmNetworkId
+  return [source, address, locationId, tokenId, subSource].filter(Boolean).join("-")
```
0xKheops: maybe this is the right opportunity to use `::` as separator instead of `-` ? 
we have some `-` in chain ids and token types, probably in other fields too, making it hard to parse when necessary

packages/balances/src/modules/EvmErc20Module.ts
```diff
@@ -98,6 +96,35 @@ export const EvmErc20Module: NewBalanceModule<
   const chainConnector = chainConnectors.evm
   assert(chainConnector, "This module requires an evm chain connector")
 
+  const prepareFetchParameters = async (
+    addressesByToken: AddressesByToken<EvmErc20Token>
+  ): Promise<EvmErc20NetworkParams> => {
+    const tokens = await chaindataProvider.tokensById()
```
0xKheops: maybe pass tokensById as parameter and make method synchronous ? would avoid fetching tokens twice when called by suscribeBalances

packages/balances/src/types/balances.ts
chidg: It's fine, although afaik the situations where we are parsing the ID are only in getting the 'placeholder balances', which will no longer exist. Are there any other reasons why the ID needs to be parseable?

0xKheops: not really, for now
it's just that it feels wrong to use a separator that's also used in the values that compose the id.
same issue in token ids actually

apps/extension/src/ui/atoms/balances.ts
```diff
@@ -24,26 +16,22 @@ import {
...
-import { logObservableUpdate } from "./utils/logObservableUpdate"
-
-const NO_OP = () => {}
 
 // Reading this atom triggers the balances backend subscription
 // Note : unsubscribing has no effect, the backend subscription will keep polling until the port (window or tab) is closed
-const rawBalancesSubscriptionAtom = atomWithSubscription<void>(
-  () => api.balances(NO_OP),
+const rawBalancesSubscriptionAtom = atomWithSubscription<BalanceSubscriptionResponse>(
+  (get) => api.balances(get),
```
alecdwm: It seems strange to me that jotai's `get` fn is being passed as a parameter to `api.balances` - is this on purpose?

chidg: This is so that the state of the atom can be updated with data from the api subscription. Is that incorrect?

packages/extension-core/src/domains/balances/pool.ts
```diff
@@ -0,0 +1,865 @@
...
+    log.log("Closing balance subscriptions")
+
+    // ignore old subscriptions if they're still closing when we next call `openSubscriptions()`
+    this.#subscriptionsGeneration = (this.#subscriptionsGeneration + 1) % Number.MAX_SAFE_INTEGER
+
+    this.#closeSubscriptionCallbacks
+      .splice(0, this.#closeSubscriptionCallbacks.length)
+      // wait 10_000ms in case the user is opening and closing the popup quickly
+      // this way the rpcs will remain connected for an extra ten seconds
+      .forEach((cb) => cb.then((close) => setTimeout(close, 10_000)))
```
chidg, line 656: @alecdwm wdyt about putting the `this.setSubscriptionsState("Closed")` in a `then` following this timeout.... that would ensure the subscriptions were actually closed when the balance modules wind down

apps/extension/src/ui/atoms/balances.ts
chidg: Looked into it again.... you were right!

packages/extension-core/src/domains/balances/pool.ts
alecdwm: I think it should stay where it is;
`setSubscriptionsState` is used to prevent two *active* sets of balance subscriptions from running at the same time.

By that, I mean that the effect of having the state be anything other than `Closed` is just that we're prevented from opening any new subscriptions.

It should have no effect on whether or not we manage to actually close the existing subscriptions when we ask to

alecdwm: The mechanism for killing off the existing subscriptions is the `subscriptionsGeneration` key. But we could replace that with an `AbortController` - which would make the code a bit tidier.

apps/extension/src/ui/atoms/balances.ts
```diff
@@ -1,20 +1,12 @@
-import { isAccountCompatibleWithChain } from "@extension/core"
-import {
-  Address,
-  Balances,
-  HydrateDb,
-  db as balancesDb,
-  deriveStatuses,
-  getValidSubscriptionIds,
-} from "@talismn/balances"
+import {} from "@extension/core"
```
alecdwm: I think this import is a typo

apps/extension/src/ui/domains/Portfolio/AssetsTable/DashboardAssetRow.tsx
```diff
@@ -142,23 +141,16 @@ export const AssetRow = ({ balances }: AssetRowProps) => {
           />
         </div>
         <div className="text-right">
-          {status.status === "initializing" ? (
-            <div className="flex h-[6.6rem]  w-full flex-col items-end justify-center gap-2 px-8">
-              <div className="bg-grey-700 rounded-xs h-8 w-[10rem] animate-pulse"></div>
-              <div className="bg-grey-700 rounded-xs h-8 w-[6rem] animate-pulse"></div>
-            </div>
-          ) : (
```
alecdwm, line 159: If I understand correctly, there has been a subtle change in what each balance status means.

I think it would be good to update the comments on `BalanceStatus` to reflect the new meanings:
```ts
// from packages/balances/src/types/balancetypes.ts

export type BalanceStatus =
  // balance is subscribed to the on-chain value and up to date
  | "live"
  // balance was retrieved from the chain but is no longer subscribed
  | "cache"
  // balance was retrieved from the chain but we're unable to create a new subscription
  | "stale"
  // balance has never been retrieved yet
  | "initializing"
```

packages/balances-react/src/util/balancesPersist.ts
```diff
@@ -0,0 +1,53 @@
...
+  localStorage.setItem("talismanBalances", deflated.toString())
+}
+
+const localStorageRetrieve: RetrieveFn = async () => {
+  const deflated = localStorage.getItem("talismanBalances")
+  if (deflated) {
+    // deflated will be a long string of numbers separated by commas
+    const deflatedArray = deflated.split(",").map((n) => parseInt(n, 10))
+    const deflatedBytes = new Uint8Array(deflatedArray.length)
+    deflatedArray.forEach((n, i) => (deflatedBytes[i] = n))
```
alecdwm, line 44: Should be able to do
```ts
const deflatedBytes = new Uint8Array(deflated.split(","))
```
for a little speed boost potential

packages/balances/src/TalismanBalancesDatabase.ts
```diff
@@ -1,25 +1,39 @@
...
+    })
+
+    this.version(4)
+      .stores({
+        balances: "id",
+        balancesBlob: "id",
+        miniMetadatas: "id, source, chainId, specName, specVersion",
+      })
+      .upgrade(upgradeBalancesDataBlob)
+
```
alecdwm: If we want to delete the `balances` table in version `5`, then we need to explicitly set `balances: null`

> Tables are not deleted unless you specify null as the stores-specification for that table in a new version.

alecdwm: Also if we intend to do that, we should test upgrading from version `3` -> `5` and make sure that the `upgradeBalancesDataBlob` upgrade still works when `5` deletes the old table.  
In theory it should run the upgrade on version 4, *then* move on to 5, but yeah we should confirm it

packages/balances/src/modules/EvmErc20Module.ts
```diff
@@ -97,9 +81,47 @@ export const EvmErc20Module: NewBalanceModule<
...
+      },
+      {} as EvmErc20NetworkParams
+    )
+  }
 
   return {
-    ...DefaultBalanceModule("evm-erc20"),
+    ...DefaultBalanceModule(moduleType),
+    get tokens() {
+      return firstValueFrom(moduleTokens)
```
alecdwm: It would be really nice to avoid using Observable.pipe and rxjs operators when we're only trying to get 1 value

In this case, `chaindataProvider.tokens()` is equivalent to calling `firstValueFrom(chaindataProvider.tokensObservable)`

Something like this would be cool:

```ts
const tokens = async () => {
  const allTokens = await chaindataProvider.tokens()
  return allTokens.filter(...).map(...)
}

return {
  ...DefaultBalanceModule(moduleType),
  get tokens,
}
```

packages/balances/src/BalanceModule.ts
```diff
@@ -147,26 +146,36 @@ interface BalanceModuleEvm<
...
 interface BalanceModuleCommon<
   TModuleType extends string,
   TTokenType extends ExtendableTokenType,
   TTransferParams extends ExtendableTransferParams
 > {
   get type(): TModuleType
 
+  /** Returns tokens of the correct type for each balance module */
+  get tokens(): Promise<Record<string, TTokenType>>
   /**
```
alecdwm: I couldn't find anywhere in the code which uses this *outside* of the balance module.
So like, `balanceModule.tokens()` inside of the balances pool or something.

If it's only meant to be used as `this.tokens` *inside* of each balance module, I think it'd be better to create it as something like `const moduleTokens` inside of the balance module, and not make it part of the public interface like this

packages/balances/src/modules/SubstrateAssetsModule.ts
```diff
@@ -454,7 +432,15 @@ async function buildQueries(
...
         const frozen = token.isFrozen || isFrozen ? amount : "0"
 
-        return new Balance({
+        const balanceValues: Array<AmountWithLabel<string>> = []
+        if (free && free > 0n)
+          balanceValues.push({ type: "free", label: "free", amount: free.toString() })
+        if (frozen && frozen > 0n)
+          balanceValues.push({ type: "locked", label: "frozen", amount: frozen.toString() })
+
+        if (balanceValues.length === 0) return null
```
alecdwm: Note to self: test that `null` correctly replaces any prior cached non-null balance

packages/balances/src/modules/SubstrateNativeModule/QueryCache.ts
```diff
@@ -0,0 +1,487 @@
...
+
+const miniMetadatasObservable = new ReplaySubject<Map<string, MiniMetadata>>(1)
+
+liveQuery(() =>
+  balancesDb.miniMetadatas.where("source").equals("substrate-native").toArray()
+).subscribe((miniMetadatas) => {
+  miniMetadatasObservable.next(
+    new Map(miniMetadatas.map((miniMetadata) => [miniMetadata.id, miniMetadata]))
+  )
+})
```
alecdwm: Probable memory leak imo

Maybe we can set up the subscription inside of QueryCache

It'll still never be shut down after init (unless we're *actually* using `QueryCache.destroy` somewhere), but at least it'll not be initialised unless we're really using the QueryCache.

alecdwm: If the liveQuery is outside of the QueryCache because we're trying to only have 1 db connection in total - instead of 1 per QueryCache instance, then we should use a variable to store that 1 db connection and only create it if it isn't already set up during `new QueryCache`.

packages/balances/src/modules/SubstrateTokensModule.ts
```diff
@@ -437,33 +409,36 @@ async function buildQueries(
...
+        const balance =
+          storageDecoder && change !== null
+            ? (storageDecoder.decode($.decodeHex(change)) as Record<
+                "free" | "reserved" | "frozen",
+                bigint
+              >)
+            : null
+
+        if (balance === null) return null
+        const { free, reserved, frozen } = balance
```
alecdwm: This is preeeetty scary, the old version would default to `0` when the substrate devs inevitably decide to change the RPC api, whereas this new version will throw when `balance` is no longer an object.

But if it's not throwing for any existing chains, I'm happy to leave it like this for now. It's unlikely to break anytime soon, and the `scale-ts` PR completely rips this decoding stuff apart in place of a better method.

packages/balances/src/modules/util/index.ts
```diff
@@ -227,8 +228,6 @@ export const deleteSubscriptionId = () => {
 }
 
 /**
- * Sets all balance statuses from `live-${string}` to either `live` or `cached`
- *
  * You should make sure that the input collection `balances` is mutable, because the statuses
  * will be changed in-place as a performance consideration.
  */
```
alecdwm: I think we can yeet this now

packages/chain-connector/src/ChainConnector.ts
```diff
@@ -170,7 +182,7 @@ export class ChainConnector {
       await this.waitForWs(ws, timeout)
     } catch (error) {
       await this.disconnectChainSocket(chainId, socketUserId)
-      throw error
+      throw new ChainConnectionError(chainId, { cause: error })
```
alecdwm, line 185: 👌

alecdwm (commented):
Looking awesome!

I have some nitpicks but the overall vibe is hell yeah

packages/balances/src/modules/SubstrateAssetsModule.ts
chidg: Missed this, should have been removed in recent changes

packages/balances/src/modules/EvmErc20Module.ts
chidg: Thanks for the explanation, previously I had `chaindataProvider.tokensById()` here but changed it because of @0xKheops comment: https://github.com/TalismanSociety/talisman/pull/1380#discussion_r155518970

It may be that the whole idea of adding a `tokens` attribute to the module type was a mistake, can scrap it if it doesn't make sense

alecdwm: Aye! (1) I think it makes sense to remove `.tokens` from the module type, but then it's still used inside the module so (2) I think it'd be good to replace this:
```ts
const moduleTokens = chaindataProvider.tokensObservable.pipe(
  map((tokens) =>
    Object.fromEntries(
      tokens
        .filter((token): token is EvmErc20Token => token.type === moduleType)
        .map((t) => [t.id, t])
    )
  )
)
...
// and then, where we need the tokens
await firstValueFrom(moduleTokens)
```
With this:
```ts
const getModuleTokens = async () =>
  util.itemsToMapById(
    (await chaindataProvider.tokens()).filter(
      (token): token is EvmErc20Token => token.type === moduleType
    )
  )
...
// and then, where we need the tokens
await getModuleTokens()
```

alecdwm: Avoids digging into rxjs territory unless we really need to subscribe to a stream of values

packages/extension-core/src/domains/balances/pool.ts
```diff
@@ -0,0 +1,875 @@
...
+      if (window.location.href !== b.location.href)
+        throw new Error(
+          `Balances pool should only be used in the background page - used in: ${window.location.href}`
+        )
+    })
+    // subscribe this store to all of the inputs it depends on
+    this.#cleanupSubs = [this.initializeChaindataSubscription()]
+
+    // if we already have subscriptions - start polling
+    if (this.#subscribers.observed) this.openSubscriptions()
```
alecdwm:
```suggestion
    // if we already have subscriptions - start polling (after chaindata init)
    if (this.#subscribers.observed) this.hasInitialised.then(() => this.openSubscriptions())
```

packages/extension-core/src/domains/balances/pool.ts
```diff
@@ -0,0 +1,875 @@
...
+    tokenId,
+    address: chainFormattedAddress,
+  }: RequestBalance): Promise<BalanceJson | undefined> {
+    assert(chainId || evmNetworkId, "chainId or evmNetworkId is required")
+
+    const address = encodeAnyAddress(chainFormattedAddress, 42)
+
+    // search for existing balance in the store
+    const storeBalances = new Balances(Object.values(this.balances))
+    const networkFilter = chainId ? { chainId } : { evmNetworkId }
```
alecdwm, line 338: If someone calls `getBalance` with just a `tokenId` and `address`, the current `networkFilter` will limit the result to only substrate chains (balances where `evmNetworkId === undefined`)

I think removing the filter for this scenario would be less surprising:
```suggestion
    const networkFilter = chainId ? { chainId } : evmNetworkId ? { evmNetworkId } : {}
```

alecdwm: Nevermind! Just noticed the `assert` at the top

apps/extension/src/ui/domains/Portfolio/AssetsTable/DashboardAssetRow.tsx
chidg: A small change in that, also there's now two types of statuses
So firstly is the status of each balance itself - there's no more `initializing` status for individual balances. A balance is only a balance once it is known to exist and have more than 0 value.
The purpose for keeping track of 'initialising' balances was to show the skeleton/shimmer element in the wallet which indicates that some balances my not yet be displayed. This functionality is now provided by the `status` value on the `BalanceSubscriptionResponse` which indicates the overall status of the balances subscription process.

packages/balances/src/modules/util/index.ts
chidg: Yeeted

packages/balances-react/src/util/balancesPersist.ts
chidg: Couldn't make this work

packages/balances/src/modules/SubstrateTokensModule.ts
chidg: See latest update :)

alecdwm (dismissed):
**L G T M 🔥🔥🔥🔥🔥**

packages/balances/src/TalismanBalancesDatabase.ts
alecdwm: It worrkksss ✨

alecdwm (approved):
🚀🚀🚀🚀🚀

PR #1384, opened 2024-03-29
fix: popup scrollbars with 125% scale

0xKheops:
tested on : 
- mises
- linux brave & FF
- windows chrome, FF and edge

also made it so scrollbars are "dark" for popup and onboarding, as they appear in case window is too small.


jonathunne (approved):
LGTM!

PR #1386, opened 2024-03-29
fix: correct stack in extension-shared console logs; drop console calls in canary/prod builds

alecdwm:
This PR fixes two issues:

1. When calling `log.log()` (imported from `packages/extension-shared/src/log.ts`) the stacktrace in the browser dev tools would always point to the `log.ts` file instead of the code which created the log.  
Now it points to the correct call site.

2. Some calls to `console.log` weren't being caught by `consoleOverride` (from `packages/extension-core/src/util/logging.ts`) in production/canary builds.  
I've replaced it with the `Terser` `drop_console` config parameter, which completely strips the console calls out from the build. Prod logs are now silent 💤


packages/extension-shared/src/log.ts
```diff
@@ -1,16 +1,26 @@
...
  **/
 export const log = {
-  error: (message: any, ...args: any[]) => DEBUG && console.error(message, ...args),
-  warn: (message: any, ...args: any[]) => DEBUG && console.warn(message, ...args),
-  log: (message: any, ...args: any[]) => DEBUG && console.log(message, ...args),
-  debug: (message: any, ...args: any[]) => DEBUG && console.debug(message, ...args),
+  error: DEBUG && typeof console.error === "function" ? console.error.bind(console) : noop,
+  warn: DEBUG && typeof console.warn === "function" ? console.warn.bind(console) : noop,
+  log: DEBUG && typeof console.log === "function" ? console.log.bind(console) : noop,
+  debug: DEBUG && typeof console.debug === "function" ? console.debug.bind(console) : noop,
```
alecdwm, line 14: Fix number 1

apps/extension/webpack/webpack.prod.js
```diff
@@ -91,7 +97,17 @@ const config = (env) => {
...
     optimization: {
       minimize: true,
-      minimizer: [new TerserPlugin({ terserOptions: { compress: true } })],
+      minimizer: [
+        new TerserPlugin({
+          terserOptions: {
+            compress: {
+              defaults: true,
+              // Drop any calls to console.error/warn/log/debug from production/canary builds, and when running tests
+              drop_console: dropConsole(env) ? ["error", "warn", "log", "debug"] : false,
```
alecdwm, line 106: Fix number 2

apps/extension/src/common/enableAnyloggerLogsInDevelopment.ts
```diff
@@ -3,4 +3,4 @@ import "anylogger-loglevel"
 import { DEBUG } from "extension-shared"
 import loglevel from "loglevel"
 
-if (DEBUG) loglevel.setLevel("info")
+loglevel.setLevel(DEBUG ? "info" : "silent")
```
alecdwm, line 6: Fix number 2

0xKheops (approved):
thx

PR #1390, opened 2024-04-08
feat: ledger accounts for asset hubs

0xKheops:
Initially filtered out invalid balances in usePortfolio but then realized it would be better to do it at the balances atom level to not worry about it ever again, hence the usePortfolio rollback at the end


apps/extension/src/ui/domains/Account/AccountAdd/AccountAddLedger/AddLedgerSelectNetwork.tsx
```diff
@@ -125,6 +125,20 @@ export const AddLedgerSelectNetwork = () => {
             <p className="text-body-secondary mt-6">
               {t("Please note: a Ledger account can only be used on a single network.")}
             </p>
+            {["polkadot", "polkadot-asset-hub"].includes(chain?.id ?? "") && (
+              <p className="text-body-secondary mt-6">
+                {t(
+                  "Additionally, a given Ledger account cannot be added in Talisman for both Polakdot and Polkadot Asset Hub as they would share the same address. Adding one would remove the other, and vice versa."
```
alecdwm: Copycheck:

`Adding the same Ledger account to both Polkadot and Polkadot Asset Hub results in them overriding each other.`

alecdwm (dismissed):
LGTM! ✨
But I think the Relay<>Asset Hub explanation could be a bit shorter

alecdwm (dismissed):
👌

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddLedger/AddLedgerSelectNetwork.tsx
```diff
@@ -125,6 +125,20 @@ export const AddLedgerSelectNetwork = () => {
             <p className="text-body-secondary mt-6">
               {t("Please note: a Ledger account can only be used on a single network.")}
             </p>
+            {["polkadot", "polkadot-asset-hub"].includes(chain?.id ?? "") && (
+              <p className="text-body-secondary mt-6">
```
chidg: Could we make this warning conditional on the 'other' account already existing?

apps/extension/src/ui/hooks/ledger/common.ts
```diff
@@ -49,22 +49,16 @@ export const ledgerNetworks = [
...
-  // Removed because no longer maintained https://support.ledger.com/hc/en-us/articles/11158925739677-Centrifuge-CFG-
-  // {
-  //   name: "centrifuge",
-  //   genesisHash: "0xb3db41421702df9a7fcac62b53ffeac85f7853cc4e689e0b93aeb3db18c09d82",
-  //   label: "Centrifuge",
-  // },
+  {
+    name: "statemine",
+    genesisHash: "0x48239ef607d7928874027a43a67689209727dfb3d3dc5e5b03a39bdc2eda771a",
+    label: "Statemine", // current name of the app in Ledger Live and on the device, hasn't been renamed to Kusama Asset Hub yet.
```
chidg, line 55: Can we label it `Statemine (Kusama Asset Hub)` ?

apps/extension/src/ui/hooks/ledger/common.ts
```diff
@@ -49,22 +49,16 @@ export const ledgerNetworks = [
...
+  },
+  {
+    name: "statemint",
+    genesisHash: "0x68d56f15f85d3136970ec16946040bc1752654e906147f7e43e9d539d7c3de2f",
+    label: "Statemint", // current name of the app in Ledger Live and on the device, hasn't been renamed to Polkadot Asset Hub yet.
```
chidg, line 60: Can we label it `Statemint (Polkadot Asset Hub)` ?

apps/extension/src/ui/atoms/balances.ts
```diff
@@ -46,12 +47,27 @@ const rawBalancesAtom = atom((get) => {
...
+  const [tokens, chains, accounts, balances] = await Promise.all([
     get(activeTokensWithTestnetsMapAtom),
+    get(activeChainsWithTestnetsMapAtom),
+    get(accountsMapAtom),
     get(rawBalancesAtom),
   ])
 
-  return balances.filter((b) => tokens[b.tokenId])
+  // exclude invalid balances
+  return balances.filter((b) => {
```
chidg, line 58: I don't understand why there was a need for a change to logic here? Isn't this filtering already done with our existing logic for ledger accounts?

Edit: oh, because we might have an account which is still the same account but is now locked to a different chain?
In that case I would prefer to just nuke the balances for that account and refresh them on the event of the account being swapped, rather than having this logic permanently in here.

chidg (changes requested):
Small questions/comments

apps/extension/src/ui/atoms/balances.ts
0xKheops: > because we might have an account which is still the same account but is now locked to a different chain?

yes that, and also i noticed that i saw tokens that shouldn't be there, 
ie : polkadot's DOT (staked balance only, to be fair) where appearing on the portfolio of a polkadot asset hub account

Feels pretty nice to have this filtering at this low level, making it unnecessary to think about it everywhere else in the UI

apps/extension/src/ui/hooks/ledger/common.ts
0xKheops: Kept it like that because on every screens that this is displayed, it will already be stated that we're interacting with Polkadot Asset Hub.
However it's important for the user to launch the ledger app called "Statemint" on his device. 
Bit concerned about the length, it could overflow in our 2 lines messages boxes

Still want me to change ?

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddLedger/AddLedgerSelectNetwork.tsx
0xKheops: ok

apps/extension/src/ui/hooks/ledger/common.ts
0xKheops: additionally, in some case we display the app name followed by the associated token symbol between parenthesis. so if we were to do that, it would display "Stamint (Polkadot Asset Hub) (DOT)"

chidg: OK. In that case, can we change the name manually only in the dropdown where the Ledger app is first chosen?

0xKheops: ok, done

chidg (approved):
LGTM

PR #1393, opened 2024-04-08
feat: avail testnet user extensions

0xKheops:
This PR allows to send funds, sign dapp txs and estimate fee on avail testnet.

However I couldn't get our extrinsic watch method to work on avail testnet.
The hash seems good as shown is this log, which is from send funds (where we actually send the tx ourselves) : 

image


alecdwm (approved):
<div align="center">

LGTM

</div>

PR #1396, opened 2024-04-09
fix: evm tx timeout

0xKheops:
This PR aims at marking timed out EVM transaction as "unknown" instead of "error", which will allow the user to access the cancel/speed-up features.

Also fixed 2 minor bugs along the way.

### DIFF
--- a/packages/extension-core/src/domains/transactions/watchEthereumTransaction.ts
+++ b/packages/extension-core/src/domains/transactions/watchEthereumTransaction.ts
@@ -58,7 +58,7 @@ export const watchEthereumTransaction = async (
         throwAfter(5 * 60_000, "Transaction not found"),
       ])
 
-      assert(receipt, "Transaction to watch not found")
+      assert(receipt, "Transaction not found")
       // check hash which may be incorrect for cancelled tx, in which case receipt includes the replacement tx hash
       if (receipt.transactionHash === hash) {
         // to test failing transactions, swap on busy AMM pools with a 0.05% slippage limit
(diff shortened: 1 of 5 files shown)


packages/extension-core/src/domains/transactions/watchEthereumTransaction.ts
```diff
@@ -77,12 +77,22 @@ export const watchEthereumTransaction = async (
         )
     } catch (err) {
       log.error("watchEthereumTransaction error: ", { err })
-      updateTransactionStatus(hash, "error")
+      const error = err as Error
+      const isNotFound = error.message === "Transaction not found"
```
alecdwm: Maybe this should be `error?.message` in case a string or something is thrown instead?

Seems unlikely, but if it happens then `error.message` will crash the UI.

alecdwm (dismissed):
LGTM!

packages/extension-core/src/domains/transactions/watchEthereumTransaction.ts
0xKheops: good catch, thanks

alecdwm (dismissed):

packages/extension-core/src/notifications/createNotification.ts
```diff
@@ -36,6 +36,17 @@ const getNotificationOptions = (
...
       }
+    case "not_found":
+      return {
+        type: "basic",
+        title: "Transaction not found",
+        message:
+          error?.shortMessage ??
+          error?.reason ??
+          error?.message ??
+          `Transaction not found on ${networkName}.`,
```
chidg: I'm not sure if this will be a meaningful message for a lot of users. How about: "We aren't able to determine the status of this transaction"

And since we know which error will have caused it, we don't need the `error?.shortMessage ??` etc in there.

chidg (changes requested):
Good fixes, but copy could be improved

chidg (approved):
LGTM

PR #1399, opened 2024-04-14
fix: spec version parsing for papi payloads

0xKheops:
Fixes specVersion property parsing from payloads created with PAPI.
Our Transaction sign UI assumed specVersion is a hex string (because that's it's only type in pjs api), but PAPI sends it as a number. This fixes the parsing of that field to handle both scenarios.

This message shouldn't appear anymore if specVersion is passed as number : 
image


chidg (approved):
LGTM

PR #1402, opened 2024-04-15
Prevent ethereum contact address being incorrectly identified as ss58 type

chidg:
Closes #1400

PR #1407, opened 2024-04-16
Add migration to clean bad contacts from address book store

chidg:
Follow on from #1402 - stored bad contacts still cause errors


0xKheops (approved):
LGTM

PR #1422, opened 2024-04-24
feat: evm tx safety

0xKheops:


apps/extension/src/ui/domains/Ethereum/useEthTransaction.ts
```diff
@@ -588,6 +588,7 @@ export const useEthTransaction = (
     networkUsage,
     setCustomSettings,
     gasSettingsByPriority,
+    // riskAnalysis,
```
chidg: Looks like this can just be deleted?

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/blowfish.ts
```diff
@@ -0,0 +1,59 @@
+import { Languages, createEvmClient } from "@blowfishxyz/api-client/v20230605"
+import { EvmNetworkId } from "@talismn/chaindata-provider"
+import { BLOWFISH_API_KEY, BLOWFISH_BASE_PATH } from "extension-shared"
+import i18next from "i18next"
+
+import { BlowfishEvmChainInfo } from "./types"
+
+// TODO pull on startup from config repo, or add these to chaindata ?
```
chidg: config repo sounds good to me

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/useRisksReview.ts
```diff
@@ -0,0 +1,18 @@
...
+import { useMemo, useState } from "react"
+import { useOpenClose } from "talisman-ui"
+
+export const useRisksReview = (action?: ActionEnum) => {
+  const [isRiskAknowledged, setIsRiskAknowledged] = useState(false)
+
+  const drawer = useOpenClose(false)
+
+  const isRiskAknowledgementRequired = useMemo(() => action === "BLOCK", [action])
+
```
chidg, line 23: Since they've already given us an Enum, it would be nice to use it rather than rely on the string

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/util.ts
```diff
@@ -0,0 +1,295 @@
+// from @blowfishxyz/ui
```
chidg: cruft?

apps/extension/webpack/SimpleLocalizeDownloadPlugin.js
```diff
@@ -51,59 +52,90 @@ module.exports = class SimpleLocalizeDownloadPlugin {
       return
     }
```
chidg, line 54: What was the problem you were trying to solve here? Seems like maybe it was downloading multiple times?

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisDrawers.tsx
```diff
@@ -0,0 +1,149 @@
...
+  )
+
+  return (
+    <div className="animate-fade-in bg-grey-850 flex w-full flex-col gap-12 rounded-t-xl p-12">
+      <div className="scrollable scrollable-700  flex-grow overflow-y-auto pr-4 text-xs  leading-[2rem]">
+        <div className="text-body-secondary leading-paragraph flex w-full flex-col gap-8">
+          <div className="text-body text-md text-center font-bold">
+            {t("Automatic risk assessments")}
+          </div>
+          <div className="text-body-secondary text-justify text-sm">
```
chidg: oops, sorry - I didn't mean justify, meant center
```suggestion
          <div className="text-body-secondary text-center text-sm">
```

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisDrawers.tsx
```diff
@@ -0,0 +1,149 @@
...
+            <p>
+              {t(
+                "Ethereum transactions and messages can be simulated on a secure server to assess their risk. Would you like to enable this feature?"
+              )}
+            </p>
+            <p className="mt-4 text-justify">
```
chidg:
```suggestion
            <p className="mt-4 text-center">
```

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/blowfish.ts
```diff
@@ -0,0 +1,59 @@
...
+export const getBlowfishChainInfo = (evmNetworkId: EvmNetworkId) => {
+  return BLOWFISH_SUPPORTED_CHAINS[evmNetworkId] ?? null
+}
+
+export const getBlowfishClient = (evmNetworkId: EvmNetworkId) => {
+  if (!BLOWFISH_SUPPORTED_CHAINS[evmNetworkId]) return null
+
+  // TODO remove this check once we have a proxy
+  // https://docs.blowfish.xyz/docs/wallet-integration-guide#optional-proxy-server
+  if (!process.env.BLOWFISH_API_KEY) return null
```
chidg, line 45: Can we remove it now?

packages/extension-shared/src/constants.ts
```diff
@@ -29,6 +29,9 @@ export const IPFS_GATEWAY = "https://talisman.mypinata.cloud/ipfs/"
 
 export const TALISMAN_CONFIG_URL = "https://talismansociety.github.io/talisman-config/config.toml"
 
+export const BLOWFISH_BASE_PATH = process.env.BLOWFISH_BASE_PATH || "https://bfp.talisman.xyz"
+export const BLOWFISH_API_KEY = process.env.BLOWFISH_API_KEY
```
chidg, line 33: I encountered some confusion when testing. I saw the above line where there is a default base path given, and assumed I did not need a base path, but I did need an API key. I got an API key from blowfish, however this setup didn't work, because our proxy is origin-limited and couldn't be accessed in dev. 

It would be good to prevent dev confusion in setting this up. Suggest that either here, or in the webpack config where we are injecting these environment variables, we should check if `BLOWFISH_API_KEY` is defined but `BLOWFISH_BASE_PATH` is not, or check if the current host is one of our whitelisted origins, and in that case raise an error warning the dev, or just use `https://api.blowfish.xyz` as the base path, or any other solution that would enable devs to get set up without having to read the blowfish docs.

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisRecommendation.tsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <div
+      className={classNames(
+        "leading-paragraph flex w-full gap-8 rounded p-4",
+        bgClassName,
+        textClassName
+      )}
+    >
+      <div>
+        <div className={classNames("rounded-full p-4", iconClassName)}>
```
chidg, line 98: Icon alignment looks a bit wacky, possible to vertically centre?
Screenshot 2024-05-01 at 1 04 05 PM

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisRecommendation.tsx
```diff
@@ -0,0 +1,117 @@
...
+        return {
+          Icon: ShieldOkIcon,
+          bgClassName: "bg-green/10",
+          textClassName: "text-green",
+          iconClassName: "bg-green/10",
+          title: result.warnings.length ? t("No Risk Found") : t("Low Risk"),
+          description:
+            result.warnings[0]?.message ??
+            (type === "transaction"
+              ? t("No risks were identified while analysing this transaction")
```
chidg:
```suggestion
              ? t("No risks were identified in this transaction")
```

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisRecommendation.tsx
```diff
@@ -0,0 +1,117 @@
...
+              : t("No risks were identified while analysing this message")),
```
chidg:
```suggestion
              : t("No risks were identified in this message")),
```

chidg:
In addition to other comments - this was when signing Typed Data v3. 
Since we know that this is an unsupported message type, can we show the 'Assessment Unavailable' message rather than a scan failed message?

Screenshot 2024-05-01 at 1 36 24 PM

chidg (changes requested):
Overall looks great and works well. Some small changes requested.

packages/extension-shared/src/constants.ts
0xKheops: yes there is a specific setup for Talisman members so they can use it in dev mode, will explain again during dev sync

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/RiskAnalysisRecommendation.tsx
0xKheops: it's 1:1 figma, top-aligned, makes even more sense when they are multiple messages with different amount of description rows

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/blowfish.ts
0xKheops: i'd leave it so people external to core team can still configure it to the regular blowfish api url

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/util.ts
0xKheops: basically wanted to say that 90% of this file is copy pasted from blowfish sources, will rephrase

apps/extension/webpack/SimpleLocalizeDownloadPlugin.js
0xKheops: yes while working in dev mode with multilanguage environment keys provided, it re-downloads on each code change, which is very annoying and felt unintended. with this change it would only download on startup

apps/extension/src/ui/domains/Sign/Ethereum/riskAnalysis/blowfish.ts
0xKheops: actually changed my mind about this: the list of chain is also hardcoded in their client library, so might want to leave it like that and remove the comment.

chidg (approved):
Love it
LGTM

PR #1428, opened 2024-05-06
fix: don't print `Setting supported languages` in dev builds

alecdwm:
We only need to check that `supportedLanguages` is configured correctly when we're building a production release.
It's always just `English` for dev.


chidg (approved):
LGTM

PR #1430, opened 2024-05-09
Restart extension automatically when clearing DB on version error

chidg:


0xKheops (commented):
~~what about doing both ?
key thing is we want the background process to restart, not just one front-end tab~~

nevermind, misread the change :)

0xKheops (approved):
LGTM

PR #1432, opened 2024-05-09
fix: enforce priority fee on opstack networks

0xKheops:
OP stack transactions will always use a minimum of 1 wei as maxPriorityPerGas

This fixes the "transaction underpriced: gas tip cap 0, minimum needed 1" error that occurs when submiting a tx on BOB.
Problem seems to be global to opstack chains, I was able to reproduce it on op sepolia.


chidg (approved):
LGTM

PR #1434, opened 2024-05-09
Remove Automata mainnet and Contextfree testnet from talisman

Liao1:


chidg:
Hi @Liao1 . Thanks for the PR. Would you mind telling us a bit about why this is needed?

Liao1:
> Hi @Liao1 . Thanks for the PR. Would you mind telling us a bit about why this is needed?

Hi @chidg , the Automata mainnet and Contextfree testnet will be facing planned maintenance and may become unavailable, so we want to remove it from the default list first.

chidg:
Thanks @Liao1 - we'll close this as it's covered in #1438

PR #1435, opened 2024-05-12
feat: uniswap v2 lp tokens

alecdwm:
- [x] Find user's LP positions via asset discovery
  - Include coingeckoIds for the underlying pool tokens
- [x] Write balance module to query LP positions
- [x] Build UI to show LP positions
- [x] Add support for fiat rates to LP tokens
- [x] Add support for custom LP tokens
- [ ] Fix SendFunds for LP tokens


packages/balances/src/modules/EvmUniswapV2Module.ts
```diff
@@ -0,0 +1,483 @@
...
+const getEvmNetworkIdFromTokenId = (tokenId: string) => {
+  const evmNetworkId = tokenId.split("-")[0] as EvmNetworkId
+  if (!evmNetworkId) throw new Error(`Can't detect chainId for token ${tokenId}`)
+  return evmNetworkId
+}
+
+export type EvmUniswapV2Token = NewTokenType<
+  ModuleType,
+  {
+    poolAddress: string
```
chidg, line 48: Is there a need to use `poolAddress` rather than the current `contractAddress` - it would be nice to maintain as much compatibility as possible with the ERC20 standard and the existing token types, then you don't have to know about what type of token it is to find out the same information.

chidg (changes requested):
Overall looks great
Feels like the send funds function is unfinished and could do with some refinement

chidg (approved):
Approving to merge and make the git gods happy
Two items outstanding:
- transfer functionality
- `poolAddress` vs. `contractAddress`

PR #1437, opened 2024-05-14
chore: updated chaindata

UrbanWill:
## ⛑️ **What was done:**

The goal is to update NeuroWeb main and test net tokens chaindata, so Signet can correctly display the token logo when using `balances` related packages.
Updated chain data after merging chaindata changes PR https://github.com/TalismanSociety/chaindata/pull/134. 

##  **Technical:**

- Ran `generate-init-data`


chidg:
H @UrbanWill ,
I'll close this as it is probably a misunderstanding - these files are really just a build artifact for the wallet itself. The changes to chaindata should be sufficient to achieve what you're trying to do. We don't actually have consensus on whether these files should even be committed to the repo as they are not really intended to be edited at all.

UrbanWill:
> H @UrbanWill , I'll close this as it is probably a misunderstanding - these files are really just a build artifact for the wallet itself. The changes to chaindata should be sufficient to achieve what you're trying to do. We don't actually have consensus on whether these files should even be committed to the repo as they are not really intended to be edited at all.

Gotcha!

PR #1471, opened 2024-06-06
Despaghettify connected accounts components

chidg:
Created a generic component for connecting Polkadot accounts which can be used in any UI context, which manages its own state regarding showing ethereum accounts and has a very simple API.


0xKheops (approved):
LGTM

PR #1473, opened 2024-06-07
feat: ledger polkadot generic app

0xKheops:


apps/playground/src/components/Substrate/balances/SendNative.tsx
```diff
@@ -0,0 +1,89 @@
...
+        const unsub = await api.tx.balances
+          .transferAllowDeath("5EXb7e8Kq9m62XTFKVYmGsHFADU4knyFNg6NKJmLKDCz4Gij", data.plancks)
+          .signAndSend(account.address, (result) => {
+            const { status } = result
+            setStatus(status)
+
+            if (status.isFinalized) {
+              unsub()
+              setTxProcessing(false)
+              // refresh()
```
chidg: needed?

0xKheops: definitely not, left over from some other component's copy paste.
thanks

packages/chaindata-provider/src/types/Chain.ts
```diff
@@ -39,6 +39,8 @@ export type Chain = {
   // TODO: Delete (has its own store now)
   /** @deprecated has its own store now */
   balancesMetadata: Array<BalancesMetadata>
+
+  hasCheckMetadataHash?: boolean // Indicates if the chain has the CheckMetadataHash extension, enabling signing with ledger generic app
```
chidg, line 43: Just curious, how do we identify if a chain has this extension, and do we need to maintain this manually in chaindata?

0xKheops: to know, need to download the metadata and see if there is a the CheckMetadataHash entry in the extrinsic's signedExtensions

It's done here automatically in chaindata :
https://github.com/TalismanSociety/chaindata/blob/feat/hasCheckMetadataHash-flag/scripts/fetch-external/steps/updateChainsExtrasCache.ts#L361

chidg (approved):
LGTM

PR #1476, opened 2024-06-10
Throw better error when Talisman not onboarded on evm dapps

chidg:
We now throw a `EthProviderRpcError` with code `ETH_ERROR_EIP1993_UNAUTHORIZED` when Talisman is not onboarded and a request is sent from an EVM dapp


0xKheops (approved):
LGTM

PR #1482, opened 2024-06-24
fix: post dedexification balances fixes 

0xKheops:
each commit fixes a different thing

### DIFF
--- a/packages/balances/src/types/balancetypes.ts
+++ b/packages/balances/src/types/balancetypes.ts
@@ -116,8 +116,11 @@ type BaseAmountWithLabel<TLabel extends string> = {
   meta?: unknown
 }
 
-export const getValueId = (amount: AmountWithLabel<string>) =>
-  [amount.label, amount.type, amount.source].join("::")
+export const getValueId = (amount: AmountWithLabel<string>) => {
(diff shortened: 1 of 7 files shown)


packages/balances/src/types/balancetypes.ts
```diff
@@ -117,9 +117,18 @@ type BaseAmountWithLabel<TLabel extends string> = {
 }
 
 export const getValueId = (amount: AmountWithLabel<string>) => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const paraId = (amount.meta as any)?.paraId ?? ""
-  return [amount.label, amount.type, amount.source, paraId].join("::")
+  const getExtra = () => {
+    const meta = amount.meta as
+      | { poolId?: number; paraId?: number; unbounding?: boolean }
```
chidg: typo: `unbounding` -> `unbonding`

chidg: Also, since unbonding is a meta data value that will only be present and true for balance values with the label `nompools-unbonding`, I don't think it should be necessary to include it as a characteristic of the ID - it's already taken care of with the inclusion of `amount.label`.

chidg (dismissed):
LGTm

chidg (approved):
LGTM

PR #1492, opened 2024-06-30
feat: migrate from `yarn` to `pnpm`

alecdwm:
Includes a fix to yeet `IToken`, which previously would cause the build to *sometimes* fail, but after migrating to `pnpm` it would fail *every* time.

Note: this warning is thrown on `pnpm install`:  
`Failed to create bin at talisman/node_modules/.bin/ledger-substrate. ENOENT: no such file or directory, open 'talisman/node_modules/@zondax/ledger-substrate/dist/cmd/cli.js'`

I had a look at the `@zondax/ledger-substrate` source code, and the `.bin` dir is no longer used in the latest version, so I think the error will fix itself when we upgrade to add support for the generic polkadot app.


.nvmrc
```diff
@@ -1 +1 @@
-v18.14.2
\ No newline at end of file
+v18.20.32
```
alecdwm, line 1: This is now the latest v18

.prettierrc.cjs
```diff
@@ -1,30 +1,43 @@
...
+
+    // local aliases / packages starting with one of these prefixes
+    "<TYPES>^(@common|@talisman|@ui|@tests|@extension/core|@extension/shared)(/.*)?$", // type imports
+    "^(@common|@talisman|@ui|@tests|@extension/core|@extension/shared)(/.*)?$", // imports
+    "", // a gap
+
+    // local `./blah/something` packages
+    "<TYPES>^[.]", // type imports
+    "^[.]", // imports
+  ],
```
alecdwm, line 27: This plugin is actively maintained (unlike the previous one), and with this config it sorts the imports in a very similar way to our previous setup, with three changes:

- It separates `import type` statements above any `import` statements
- It uses a case-insensitive sort (so AbcDefG instead of ADGbcef)
- It lets us put our `@common`/`@ui`/etc aliases next to our `./relative` imports

apps/extension/jest.config.js
```diff
@@ -6,7 +6,7 @@ const { compilerOptions } = require("./tsconfig.json")
 
 module.exports = {
   transformIgnorePatterns: [
-    "/node_modules/(?!@polkadot|@substrate|@azns|@babel/runtime/helpers/esm/|@metamask|url-join|isows|nanoid)",
+    "/node_modules/.pnpm/(?!@polkadot|@substrate|@azns|@babel/runtime/helpers/esm/|@metamask|url-join|isows|nanoid|@talismn|scale-ts)",
```
alecdwm, line 9: Because of pnpm's [isolated](https://pnpm.io/npmrc#node-linker) `node_modules` structure, jest finds all of the dependencies inside of `node_modules/.pnpm`

package.json
```diff
@@ -45,91 +48,86 @@
...
-    "@substrate/txwrapper-examples": "7.2.0",
-    "@substrate/txwrapper-orml": "7.2.0",
-    "@substrate/txwrapper-polkadot": "7.2.0",
-    "@substrate/txwrapper-registry": "7.2.0",
-    "@substrate/txwrapper-substrate": "7.2.0",
-    "@substrate/txwrapper-template": "7.2.0",
-    "glob-parent": "^5.1.2",
-    "decode-uri-component": "0.2.1",
-    "semver": "~7.5.4"
+    "turbo": "1.3.1"
```
alecdwm, line 52: I don't know why this was `latest` before, but I've added the version which was in our `yarn.lock` to this `package.json`

package.json
```diff
@@ -45,91 +48,86 @@
...
   },
-  "engines": {
-    "node": ">=18"
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "@sentry/cli"
+    ],
```
alecdwm, line 57: I've tested this to 100% confirm that **only** dependencies listed here will have their postinstall scripts executed.

If we ditch `@sentry/cli`, or don't want to run *any* postinstall scripts, we can delete this section from the `package.json` and add [`ignore-scripts=true`](https://pnpm.io/npmrc#ignore-scripts) or [`ignore-dep-scripts=true`](https://pnpm.io/npmrc#ignore-dep-scripts) to `.npmrc`

turbo.json
```diff
@@ -1,15 +1,6 @@
...
   "pipeline": {
-    "postinstall": {
-      "cache": false
-    },
-    "dev": {
-      "cache": false
-    },
-    "update:translations": {
-      "outputs": ["public/locales/**"]
-    },
```
alecdwm, line 12: We only use `turbo` for our `build` commands, the other ones don't need it!

chidg (approved):
LGTM

PR #1499, opened 2024-07-02
chore: fix sentry in the backend

chidg:
Replace calls to `Sentry.captureException` with our own homebrewed method which calls the sentry scope directly, this is needed due to the custom configuration in the service worker.


0xKheops (approved):
LGTM

PR #1505, opened 2024-07-03
fix: connection issues from extension frontend in firefox

chidg:
Enables frontend messages to wake up background script in FF
Does not fix similar problem from dapp side :(


0xKheops (approved):
LGTM!

0xKheops (approved):

PR #1511, opened 2024-07-04
fix: use Error type for cause of `Unable to complete action` error

alecdwm:
A fix for https://talisman.sentry.io/issues/5570357350 and https://talisman.sentry.io/issues/5568629235


chidg (approved):
LGTM!

PR #1514, opened 2024-07-04
fix: expected errors spamming sentry

alecdwm:
Fixes these, which are spamming sentry atm:
- https://talisman.sentry.io/issues/5568293840
- https://talisman.sentry.io/issues/5568360495
- https://talisman.sentry.io/issues/5568153931
- https://talisman.sentry.io/issues/5568844920
- https://talisman.sentry.io/issues/5568309829
- https://talisman.sentry.io/issues/5568310235

### DIFF
--- a/packages/extension-core/src/domains/sitesAuthorised/store.ts
+++ b/packages/extension-core/src/domains/sitesAuthorised/store.ts
@@ -1,4 +1,3 @@
-import { assert } from "@polkadot/util"
 import { convertAddress } from "@talismn/util"
 
 import { SubscribableByIdStorageProvider } from "../../libs/Store"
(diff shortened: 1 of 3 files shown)


chidg (commented):
Not sure about this method of doing this
Our sentry config already comes with an array of `ignoreErrors` regexes. I think it's preferably to continue with that method rather than creating a new way of doing something similar.

alecdwm:
> Not sure about this method of doing this Our sentry config already comes with an array of `ignoreErrors` regexes. I think it's preferably to continue with that method rather than creating a new way of doing something similar.

Yeah, that definitely makes sense!  
I do wish there was a better built-in way for sentry exclusions than just regex matching on the error message. I'm a big fan of these `instanceof` checks 😭  
But you're right they're not worth the cost of yet another method to exclude errors

PR #1515, opened 2024-07-04
fix: crash in requests handler

alecdwm:
Fixes https://talisman.sentry.io/issues/5571043034


chidg:
I'm leaning towards *not* merging this at the moment, because I think it might be a side effect of other bugs related to manifest v3. Let's leave it open and see what happens with the next release.

PR #1519, opened 2024-07-05
fix: don't expect networks and chains lists to have entries

0xKheops:
removed a check on evmNetworks and chains to both have entries before displaying tokens, which is useless since we now have a suspense on those atoms

It fixes a bug where tokens list wouldn't display if all EVM networks are disabled, or if all Polkadot networks are disabled.


chidg (approved):
LGTM

PR #1523, opened 2024-07-05
feat: connection status component on ledger generic add account screen

### LINKED ISSUE #1524: Issues importing Polkadot ledger accounts with specific address index
**Describe the bug**
I'm trying to import Polkadot accounts from ledger that have the account type (first derivation level) set to 0 but the account index (second derivation level) set to different values. They don't show up in the import list, is it possible to import them?

**Screenshots**
This is how I imported the accounts in polkadot-js, note that the second selector has a non-zero value.

Screenshot from 2024-07-06 19-58-41

0xKheops:
less than 2 minutes :)


chidg (approved):
LGTM

Issue #918, opened 2023-07-09
Polkadot Vault ethereum accounts support


0xKheops:
PV seems to be compatible with moonbeam/moonriver, but Talisman sees QR code as invalid

alecdwm:
The `invalid qr code` error has been fixed, but we still can't sign Moonbeam txs with PV.

However, it also doesn't seem to work when using PV with PJS apps:

1. Add Moonbeam chainspec & metadata to PV via https://metadata.novasama.io
2. Create Moonbeam account in PV
3. Connect Moonbeam account to https://polkadot.js.org/apps/
4. Prepare a transfer TX to be signed by the PV account
5. Scan the QR code with the vault, and note the `Please Add The Network You Want To Transact in` error

<div align="center">
<img src="https://github.com/user-attachments/assets/7b9dd75b-a1f8-4e8e-8bf8-567998bfd7c5" width="33%" />
</div>

chidg:
Closed by #1535

PR #1535, opened 2024-07-16
fix: qr scanning on firefox

### LINKED ISSUE #1275: Users with multiple webcams cannot select which one to use for QR code scanning (Polkadot Vault)
It should be possible to get a list of video inputs using the [enumerateDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices) api.

With that list, we can show a dropdown when `# of video inputs > 1`.

If the user selects a device with the dropdown, we can use the `deviceId` parameter to `getUserMedia`:
```ts
getUserMedia({
  video: {
    deviceId: {
      exact: myExactCameraOrBustDeviceId,
    },
  },
})
```

The tricky part, is that our call to `getUserMedia` is nested inside of `react-qr-reader`, which in turn is nested inside of `@polkadot/react-qr`.

So, we need to fork [`react-qr-reader`](https://github.com/JodusNodus/react-qr-reader) to add a `deviceId` property to its [QrReader](https://github.com/JodusNodus/react-qr-reader?tab=readme-ov-file#component-api) component (which is imported as just `Reader` inside of `@polkadot/react-qr`).

Then, we need to fork `@polkadot/react-qr` to expose this new property on the [`QrScanAddress`](https://github.com/polkadot-js/ui/blob/master/packages/react-qr/src/ScanAddress.tsx) and [`QrScanSignature`](https://github.com/polkadot-js/ui/blob/master/packages/react-qr/src/ScanSignature.tsx) components.

Maybe we can even get our forks PR'd back upstream?

### LINKED ISSUE #1533: Polkadot Vault no longer working
In particular Talisman now just fails to scan the QR code of the signer. I can hold it there forever, change display brightness, go closer, farther. Nothing works. This used to work completely flawlessly.

alecdwm:
Fixes:
1. https://github.com/TalismanSociety/talisman/issues/1533
  The solution is to replace our qr-scanning library jsqr which internally calls the following code, with [@zxing/browser](https://www.npmjs.com/package/@zxing/browser) which does not:
    ```ts
    new Worker(URL.createObjectURL(new Blob(["<arbitrary-js-here>"], { type: "application/javascript" })))
    ```
1. https://github.com/TalismanSociety/talisman/issues/1275
1. https://github.com/TalismanSociety/talisman/issues/918 (not quite, but makes progress towards a fix)

---

<details>
<summary>Some more details on the `new Blob` stuff</summary>

### Manifest V3 introduces some changes to the allowed CSP

- https://developer.chrome.com/docs/extensions/develop/migrate/improve-security#remove-unsupported-csv
- the relevant CSP directive to allow [jsqr](https://github.com/cozmo/jsQR) to run would be `worker-src 'self' blob:;`
- `blob:` isn't in the allowed list for MV3
- despite `blob:` not being in the CSP, chrome still lets the code pass.
- Firefox OTOH, rejects the code as the spec says it should.

#### Chrome example with `blob:` and `data:` (inline script via blob is executed despite the CSP preventing them)
image

#### Firefox example with `blob:` and `data:` (inline script via blob is rejected)
image

</details>

<details>
<summary>Dependency changes for QR code scanning</summary>

### Previous dependency tree for QR code scanning

```
└─ @polkadot/react-qr@npm:3.6.4
   └─ react-qr-reader@npm:2.2.1
      └─ jsqr@npm:1.4.0
```

NOTE: Starting with version 3, `react-qr-reader` was refactored to use `@zxing/browser` some years ago.  
However, the refactor was never released to npm.

### New dependency tree for QR code scanning
```
└─ @zxing/browser@npm:0.1.5
```

</details>


chidg (approved):
LGTM

PR #1536, opened 2024-07-16
fix: dashboard accounts fiat value wrapping bug

alecdwm:
<div align="center">

**before**
Screenshot 2024-07-16 at 12 33 52

**after**
Screenshot 2024-07-16 at 12 35 00

</div>


chidg (approved):
LGTM

PR #1539, opened 2024-07-19
fix: extension build performance issues

alecdwm:
<div align="center">

ludicrous speed, go!

</div>

### List of changes made by this PR:
**High performance impact**
- Tailwind parses our source files to determine which tailwind classes are in use by our project. It was previously parsing our entire node_modules directory, but we only want it to parse our own source files.
- Moved `CircularDependencyPlugin` from common webpack config to prod webpack config.
- Disabled sourceMaps for css.

**Medium performance impact**
- Disabled `import` parsing by css-loader (because it's already handled by postcss-import)
- Disabled `url` parsing by css-loader, which was exclusively used for webpacking fonts, and moved all fonts to the `public` dir instead.
- Disabled postcss for css files imported via node_modules
- Switched from `ts-loader` to `esbuild-loader` for typescript -> javascript transpilation
- Added node_modules exclusions to all webpack loaders which we are only using for our own source files.
- Switch large `png-inside-of-svg` assets for `proper svg` assets
- Ran largest svg files through `svgo` to reduce their filesize.
- Switched `What's New` quests banner from a react component svg to a url-imported svg.

**No performance impact**
- Added `talisman-ui` to changesets-ignored packages (as it is a private package)
- Tweaked postcss configs to follow the style used in the tailwind docs
- Moved extension dev dependencies into the devDependencies section of the package.json
- `EsbuildPlugin` is an alternative to `DefinePlugin` which we may or may not use in the future. If we do use it, it'll now be compatible with our `SimpleLocalizeDownloadPlugin`, which hooks into EsbuildPlugin/DefinePlugin to set `process.env.SUPPORTED_LANGUAGES`.
- Added `TsconfigPathsPlugin` to automatically add our tsconfig `paths` to webpack's `resolve.alias` config.
- Set webpack to `minimal` stats output by default.
- Added webpack.ProgressPlugin for non-CI builds
- Added `SpeedMeasurePlugin`, disabled by default but can be enabled via `MEASURE_WEBPACK_SPEED=true pnpm dev:extension`
- Set env.build to `dev` for dev, instead of `undefined`.
- Upgraded caniuse-lite db, and forced all copies of it to latest version using pnpm.overrides


apps/extension/tailwind.config.cjs
```diff
@@ -5,7 +5,7 @@ const TALISMAN_TAILWIND_CONFIG = require("talisman-ui/tailwind.config.cjs")
 module.exports = {
   ...TALISMAN_TAILWIND_CONFIG,
   content: [
-    "./**/*.{html,ts,tsx,svg}",
+    "./src/**/*.{html,ts,tsx,svg}",
```
alecdwm, line 8: <div align="center">

pacha-perfect

</div>

0xKheops (commented):
WILD perf boost there, thank you so much!! ❤️ 

Fonts are missing on playground (and i assume on balances-demo if used there, idk)
Other than that, LGTM !!

Leaving final approval to @chidg

packages/extension-core/src/config/posthog.ts
```diff
@@ -27,7 +27,7 @@ const unsafeProperties = [
 const talismanProperties = {
   appVersion: process.env.VERSION,
   appBuild: process.env.BUILD,
-  testBuild: DEBUG || ["qa", "ci"].includes(process.env.BUILD as string),
+  testBuild: DEBUG || ["dev", "qa", "ci"].includes(process.env.BUILD as string),
```
chidg, line 30: Dev builds shouldn't have a POSTHOG_AUTH_TOKEN set. Is it a good idea to have this here?

alecdwm: I added it because I'd expect that if for some reason someone did set an auth token in dev, we'd want those events labelled as test events not prod events

alecdwm: But FWIW, `DEBUG` is true unless `process.env.BUILD` is either `production` or `canary`

alecdwm:
#### `dev` measurements
initial build (dev)
Screenshot 2024-07-22 at 14 04 47
hot reload
Screenshot 2024-07-22 at 14 05 30

#### `fix/extension-build-performance` measurements
initial build (dev)
Screenshot 2024-07-22 at 14 15 02
hot reload
Screenshot 2024-07-22 at 14 17 46

packages/extension-core/src/config/posthog.ts
chidg: ok, seems like it won't cause any harm

chidg (approved):
LGTM

PR #1553, opened 2024-07-29
fix: don't create liveQuery subscription in SubNativeModule until needed

alecdwm:
Fixes a bug where attempting to `fetch-external` on the chaindata repo will throw an error about the IndexedDb api missing


0xKheops (approved):
LGTM :)

PR #1554, opened 2024-07-29
feat: gas mode detection improvements

### LINKED ISSUE #1529: Invalid gas settings on scroll network
Transactions remain pending due to invalid gas settings on Scroll network.

Gas mode detection seems broken as that network expects pre-eip-1559 gas settings, but Talisman provides the eip-1559 gas settings UI, 

Also it would benefit from a contract call to compute L1 Fee and display those the same way as we do for OPStack networks.

https://docs.scroll.io/en/developers/transaction-fees-on-scroll/

0xKheops:
- Fixes #1529


apps/extension/src/ui/domains/Ethereum/useEthTransaction.ts
```diff
@@ -345,7 +346,7 @@ const useGasSettings = ({
           mapMaxPriority.high = minimumMaxPriorityFeePerGas
       }
 
-      if (evmNetworkId && isOpStackEvmNetwork(evmNetworkId)) {
+      if (evmNetworkId && evmNetwork?.l2FeeType?.type === "op-stack") {
```
chidg, line 349: Is there a problem here for OP stack custom networks that are not defined in chaindata?

0xKheops: most likely yes

chidg (approved):
LGTM

PR #1555, opened 2024-07-30
fix: signing on chains with empty system properties

0xKheops:
Allows signing on local chains, ie: `./polkadot --dev`

It will sign the old way, without adding a metadata hash on the payload


alecdwm (approved):
lgtm

PR #1565, opened 2024-08-02
fix: qr code scanning broken in production builds

alecdwm:
The QR code reader lib we're using ([ZXing](https://www.npmjs.com/package/@zxing/browser)) uses [ts-custom-error](https://www.npmjs.com/package/ts-custom-error#minification-and-transpilation-mangle-custom-error-names) to define its custom `Error` types.

A known issue with this lib is that the error names are mangled by terser/uglify/other code minifiers: 
[#minification-and-transpilation-mangle-custom-error-names](https://www.npmjs.com/package/ts-custom-error#minification-and-transpilation-mangle-custom-error-names)

This broke our detection of the `NotFoundException` error, which we need to ignore in order to continue scanning a video feed which doesn't immediately have a QR code visible in the frame.

### DIFF
--- a/apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/Scan.tsx
+++ b/apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/Scan.tsx
@@ -1,8 +1,9 @@
+import { ExternalLinkIcon } from "@talismn/icons"
+import { useTranslation } from "react-i18next"
+
 import { POLKADOT_VAULT_DOCS_URL } from "@extension/shared"
 import { HeaderBlock } from "@talisman/components/HeaderBlock"
-import { ExternalLinkIcon } from "@talismn/icons"
 import { ScanQr } from "@ui/domains/Sign/Qr/ScanQr"
-import { useTranslation } from "react-i18next"
 
 import { useAccountAddQr } from "./context"
 
@@ -106,26 +107,29 @@ export const Scan = () => {
             enable={state.enable}
             error={!!state.cameraError}
             onScan={(scanned) => dispatch({ method: "onScan", scanned })}
-            onError={(error) =>
-              [
+            onError={(error) => {
+              const cameraErrors = [
                 "AbortError",
                 "NotAllowedError",
                 "NotFoundError",
                 "NotReadableError",
                 "OverconstrainedError",
                 "SecurityError",
-              ].includes(error.name)
-                ? dispatch({
-                    method: "setCameraError",
-                    error: error.name ?? error.message ?? "error",
-                  })
-                : dispatch({
-                    method: "setScanError",
-                    error: error.message.startsWith("Invalid prefix received")
-                      ? t("QR code is not valid")
-                      : error.message ?? "Unknown error",
-                  })
-            }
+              ]
+              if (cameraErrors.includes(error.name))
+                return dispatch({
+                  method: "setCameraError",
+                  error: error.name ?? error.message ?? "error",
+                })
+
+              dispatch({
+                method: "setScanError",
+                error: error.message.startsWith("Invalid prefix received")
+                  ? t("QR code is not valid")
+                  : error.message ?? "Unknown error",
+              })
+              console.error("QR code scanning error", error) // eslint-disable-line no-console
+            }}
           />
           {state.scanError && (
             <div className="text-alert-error bg-alert-error/10 mt-6 inline-block w-[260px] rounded p-4 text-center text-xs font-light">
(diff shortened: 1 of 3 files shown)


apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/Scan.tsx
```diff
@@ -106,26 +107,29 @@ export const Scan = () => {
...
+                  error: error.name ?? error.message ?? "error",
+                })
+
+              dispatch({
+                method: "setScanError",
+                error: error.message.startsWith("Invalid prefix received")
+                  ? t("QR code is not valid")
+                  : error.message ?? "Unknown error",
+              })
+              console.error("QR code scanning error", error) // eslint-disable-line no-console
```
alecdwm, line 131: Added this to make debugging similar problems easier in the future.

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/context.tsx
```diff
@@ -80,7 +78,7 @@ export const reducer = (state: AddQrState, action: Action): AddQrState => {
 
       if (!scanned) return state
       if (!scanned.isAddress)
-        return { type: "SCAN", enable: true, scanError: "QR code is not valid" }
+        return { type: "SCAN", enable: true, scanError: "QR code does not contain an address" }
```
alecdwm, line 81: Changed this so that in the future it won't be confused with the other `QR code is not valid` error.

apps/extension/src/ui/domains/Sign/Qr/ScanQr.tsx
```diff
@@ -109,30 +110,38 @@ const Scanner = ({
...
+          aborted.signal.addEventListener("abort", () => controls.stop())
+          abortedSignalSet = true
+        }
         if (aborted.signal.aborted) return
-        if (error?.name === "NotFoundException") return
 
-        if (error) onError?.(new Error(error.name, { cause: error }))
+        if (error instanceof NotFoundException) return
+        if (error instanceof ChecksumException) return
+        if (error instanceof FormatException) return
```
alecdwm, line 131: `instanceof` won't break on minified builds, unlike the previous `error?.name ===` test

apps/extension/src/ui/domains/Sign/Qr/ScanQr.tsx
```diff
@@ -109,30 +110,38 @@ const Scanner = ({
...
 
+  const codeReader = useAtomValue(codeReaderAtom)
+
   useEffect(() => {
+    if (!codeReader) return
     if (!preview.current) return
 
     const aborted = new AbortController()
-
-    const codeReader = new BrowserQRCodeReader()
```
alecdwm, line 117: No need to create a new one of these each time we run this effect, we can just use the one from the `codeReaderAtom`

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/Scan.tsx
0xKheops: might want to use `log.error` then :)

apps/extension/src/ui/domains/Sign/Qr/ScanQr.tsx
```diff
@@ -109,30 +110,38 @@ const Scanner = ({
...
-
+    let abortedSignalSet = false
     codeReader
       .decodeFromVideoDevice(selectedVideoInput, preview.current, (result, error, controls) => {
+        if (!abortedSignalSet) {
+          aborted.signal.addEventListener("abort", () => controls.stop())
```
0xKheops: Might want to remove the listener in the return function of the useEffect, to prevent a memory leak

0xKheops (changes requested):
Looking good, just need a fix for the mem leak

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddQr/Scan.tsx
0xKheops: nevermind, I thought log.error would ignore the DEBUG flag, but it doesn't

apps/extension/src/ui/domains/Sign/Qr/ScanQr.tsx
alecdwm: Nice catch!

We call `aborted.abort()` in the useEffect teardown fn, and then this listener now removes itself thanks to `{ once: true }`

0xKheops (approved):
LGTM

PR #1566, opened 2024-08-02
fix: balances initializing status

0xKheops:
The intializing status is now derived from 2 observables : 
- the array of balances that are initializing (as before)
- a new flag indicating that a subscription restart is pending


chidg (approved):
Very nice
LGTM!

PR #1573, opened 2024-08-06
fix: analytics improvements

alecdwm:
- Improved privacy-rounding function
- Removed unnecessary subscription on `appStore` in `Analytics` class
- Decoupled `GeneralReport` logic out from the core `Analytics` class
- Added new analytics props:
  - `watchedAccountsCount`
  - `lastGeneralReport`
- Fixed `GeneralReport` bugs & edge cases:
  - Include all account types in accountBreakdown (including types not included in `AccountType` enum)
  - Fetch async data in parallel instead of in sequence
  - Don't run sync processing as async tasks
  - Include a real error message when db access fails

---

Some more issues I found during testing:
- Each restart of the wallet generates a new user UUID, which leads to double-counting of assets ([fixed](https://github.com/TalismanSociety/talisman/pull/1573/commits/54bc375047f8df9ef9388adc1247ad18a683a181))
- GeneralReports are be generated from the current state of the balances pool, regardless of whether or not it is ready or still initialising ([fixed](https://github.com/TalismanSociety/talisman/pull/1573/commits/f8d0254361718094c9ed24750a062a4557ba818b))
- Same for NFTs

### DIFF
--- a/packages/extension-core/src/libs/Analytics.ts
+++ b/packages/extension-core/src/libs/Analytics.ts
@@ -1,242 +1,53 @@
-import keyring from "@polkadot/ui-keyring"
+import type { Properties } from "posthog-js"
 import { DEBUG } from "extension-shared"
-import posthog, { Properties } from "posthog-js"
+import posthog from "posthog-js"
 
-import { initPosthog } from "../config/posthog"
-import { db } from "../db"
-import { AccountType } from "../domains/accounts/types"
-import { appStore } from "../domains/app/store.app"
+import { initPosthog, setPosthogOptInPreference } from "../config/posthog"
 import { settingsStore } from "../domains/app/store.settings"
-import { balancePool } from "../domains/balances/pool"
-import { Balance, Balances } from "../domains/balances/types"
-import { getNftCollectionFloorUsd } from "../domains/nfts"
-import { nftsStore$ } from "../domains/nfts/store"
-import { chaindataProvider } from "../rpcs/chaindata"
-import { hasGhostsOfThePast } from "../util/hasGhostsOfThePast"
-import { roundToFirstInteger } from "../util/roundToFirstInteger"
-
-const REPORTING_PERIOD = 24 * 3600 * 1000 // 24 hours
-
-const ensurePosthogPreferences = (useAnalyticsTracking: boolean | undefined) => {
-  if (useAnalyticsTracking === undefined) {
-    if (posthog.has_opted_in_capturing() || posthog.has_opted_out_capturing())
-      posthog.clear_opt_in_out_capturing()
-  } else if (
-    useAnalyticsTracking &&
-    (!posthog.has_opted_in_capturing() || posthog.has_opted_out_capturing())
-  ) {
-    posthog.opt_in_capturing()
-  } else if (
-    !useAnalyticsTracking &&
-    (posthog.has_opted_in_capturing() || !posthog.has_opted_out_capturing())
-  ) {
-    posthog.opt_out_capturing()
-  }
-}
+import { withGeneralReport } from "./GeneralReport"
 
 class TalismanAnalytics {
-  lastGeneralReport: undefined | number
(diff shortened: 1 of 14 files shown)


packages/extension-core/src/libs/Analytics.ts
```diff
@@ -1,242 +1,53 @@
...
-import { Balance, Balances } from "../domains/balances/types"
-import { getNftCollectionFloorUsd } from "../domains/nfts"
-import { nftsStore$ } from "../domains/nfts/store"
-import { chaindataProvider } from "../rpcs/chaindata"
-import { hasGhostsOfThePast } from "../util/hasGhostsOfThePast"
-import { roundToFirstInteger } from "../util/roundToFirstInteger"
-
-const REPORTING_PERIOD = 24 * 3600 * 1000 // 24 hours
-
-const ensurePosthogPreferences = (useAnalyticsTracking: boolean | undefined) => {
```
chidg, line 20: ~~This was needed to get around some buggy behaviour in Posthog, not sure if safe to remove~~
Sorry, I saw you've replaced it with a nicer version elsewhere

chidg (dismissed):
LGTM

packages/extension-core/src/libs/GeneralReport.ts
```diff
@@ -59,6 +68,31 @@ async function getGeneralReport() {
...
+  if (balancePool.hasInitialised.isPending()) return
+  // if balancePool isn't initialised (all balance queries set to `live`), balances stats may be incorrect (initialising state 2)
+  if (poolStatus === "initialising") return
+
+  // if the balancePool has cached balances, we probably have an incomplete state of the user's balances
+  // (i.e. we're still fetching them after a browser restart)
+  //
+  // 30 seconds after the pool is initialised, all of the cached balances are transformed into stale balances
+  //
+  // in the future, we should expose an observable on the pool which can indicate whether or not the pool is hydrated,
```
chidg, line 89: I think this was done in Kheops' recent work?

chidg (approved):
LGTM!
one comment not blocking

packages/extension-core/src/libs/GeneralReport.ts
alecdwm: I think Kheops' new `isRestartPending` state covers this edge-case for code which subscribes to balances, because creating a subscription sets `isRestartPending` to `true`.  
But in the case of analytics here, we're not telling the pool to subscribe to balances for us, we're just asking for the pool's current state and hoping someone else has set up the subscription.

As far as I can tell, this means we might ask the pool for its current state when neither of these two are true:
- `isRestartPending`
- `initialising.length`

alecdwm: So I think we still need this check here to avoid the edge-case where we try and do a general report on a pool which hasn't been subscribed to since the last background script restart

chidg: ok - I was thinking of the `#balances` property on the `BalancesPool`, which returns `{status: "initialising"}` and abstracts away the `#initialising` property of the pool - see https://github.com/TalismanSociety/talisman/blob/6d68f92adbd98bdea3699331d2742105ba545508/packages/extension-core/src/domains/balances/pool.ts#L124

and then:
https://github.com/TalismanSociety/talisman/blob/6d68f92adbd98bdea3699331d2742105ba545508/packages/extension-core/src/domains/balances/pool.ts#L159
Where those two properties you mention are combined

alecdwm: I think the semantics of `#balances` is:

- init background script
  - `status: "live"`
- create first balances subscription
  - `status: "initialising"`
     (at first due to `isRestartPending`, and then due to a non-0-length list of initialising balances)
- all balances initialised
  - `status: "live"`

chidg: Correct, I tested it and confirmed! LGTM then

PR #1578, opened 2024-08-09
fix: allow signing when `Extrinsic` type is invalid

alecdwm:
Workaround for https://github.com/freeverseio/laos/issues/710
Fixes signing, at the expense of no ledger support or tx watching in the wallet

### DIFF
--- a/packages/extension-core/src/domains/signing/handler.ts
+++ b/packages/extension-core/src/domains/signing/handler.ts
@@ -13,6 +13,7 @@ import type {
   RequestSigningApproveSignature,
   SignerPayloadJSON,
 } from "./types"
+import { sentry } from "../../config/sentry"
 import { getPairForAddressSafely } from "../../handlers/helpers"
 import { talismanAnalytics } from "../../libs/Analytics"
 import { ExtensionHandler } from "../../libs/Handler"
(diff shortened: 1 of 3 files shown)


packages/extension-core/src/domains/signing/handler.ts
```diff
@@ -94,16 +95,22 @@ export default class SigningHandler extends ExtensionHandler {
...
-          signedTransaction = tx.toHex()
+          try {
+            const tx = registry.createType(
+              "Extrinsic",
+              { method: payload.method },
+              { version: payload.version }
+            )
+
+            // apply signature to the modified payload
+            tx.addSignature(payload.address, signature, payload)
```
alecdwm, line 106: The v15 chain metadata instructs us to build a 66-byte signature, but the chain will reject anything other than a 65-byte signature.

We trim the first byte off of the signature to enable signing.
However, `registry` still expects to see a 66-byte signature (as its understanding is based on the v15 chain metadata).

As such, our workaround is the `try {} catch {}` around this block which will *attempt* to create the `signedTransaction` needed for ledger support, but will gracefully fall back if it cannot.

packages/extension-core/src/domains/transactions/watchSubstrateTransaction.ts
```diff
@@ -259,22 +259,25 @@ export const watchSubstrateTransaction = async (
 
   assert(chain.genesisHash === payload.genesisHash, "Genesis hash mismatch")
 
-  const hash = getExtrinsicHash(registry, payload, signature)
+  try {
+    const hash = getExtrinsicHash(registry, payload, signature)
```
alecdwm, line 263: Same as above, but for tx watching

0xKheops (approved):
LGTM

PR #1583, opened 2024-08-13
feat: ledger onboarding rework

0xKheops:
Select network first, so we can guide user to the correct ledger app type


apps/extension/src/ui/hooks/ledger/useLedgerSubstrateChains.ts
```diff
@@ -0,0 +1,50 @@
...
+import useChains from "../useChains"
+import { CHAIN_ID_TO_LEDGER_APP_NAME } from "./common"
+
+export type ChainWithLedgerApp = Chain & {
+  ledgerAppName?: string
+  supportedLedgerApps: AddSubstrateLedgerAppType[]
+}
+
+export const useLedgerSubstrateChains = () => {
+  const { chains } = useChains({ includeTestnets: false, activeOnly: true })
```
chidg: wdyt about allowing the choice of all chains, not `activeOnly` ? 
I just had the problem where I have deactivated some chains but forgotten about it, and was confused why they were missing from the list.

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddLedger/AddLedgerSelectNetwork.tsx
```diff
@@ -303,47 +195,61 @@ export const AddLedgerSelectNetwork = () => {
...
-
-            {substrateAppType === AddSubstrateLedgerAppType.Legacy && (
-              <SubstrateLegacyNetworkSelect chain={chain} onChange={handleNetworkChange} />
-            )}
-            {substrateAppType === AddSubstrateLedgerAppType.Migration && (
-              <SubstrateMigrationNetworkSelect
-                app={migrationApp}
-                onChange={handleMigrationAppChange}
-              />
+            {!!chain && (
```
chidg, line 207: There's a slightly annoying height-shift in this component when selecting different networks because some of them have longer names, which forces the text to be 3 lines instead of 2

apps/extension/src/ui/hooks/ledger/useLedgerSubstrateChains.ts
0xKheops: sure, will change it

apps/extension/src/ui/domains/Account/AccountAdd/AccountAddLedger/AddLedgerSelectNetwork.tsx
0xKheops: ok will fix

chidg (approved):
LGTM

PR #1584, opened 2024-08-15
fix: run terser in 1 thread and use swc minify

alecdwm:


chidg (approved):
Changed base to 1.30.1
LGTM!

PR #1598, opened 2024-09-04
fix: make build:review compatible with 🥔s and 🧮s

alecdwm:
Previously:
- Use docker buildx to install deps & build the `.zip` file for the release

Now:
- Use docker build (legacy builder) to install deps to an image, then use docker run a container using the image (with ./review bind-mounted to the host) to build the `.zip` file for the release


chidg (approved):
LGTM

PR #1599, opened 2024-09-05
feat: rpc priority rotation & resets

0xKheops:
- Manages RPCs priority with a list that can be rotated
- Rotates RPCs and restarts the websocket in case of timeout for a request (30s) or if a rate limit error is detected
- Editing or reseting a network (from settings) will reset the current websocket and db entry to ensure new RPCs list is used

In case a reset currently pending requests will fail, but subsequent ones should use the next rpc in the list

### DIFF
--- a/packages/connection-meta/src/TalismanConnectionMetaDatabase.ts
+++ b/packages/connection-meta/src/TalismanConnectionMetaDatabase.ts
@@ -25,6 +25,7 @@ export class TalismanConnectionMetaDatabase extends Dexie {
       //
       // Never index properties containing images, movies or large (huge) strings. Store them in IndexedDB, yes! but just don’t index them!
       // https://dexie.org/docs/Version/Version.stores()#warning
+      chainPriorityRpc: null, // delete legacy table
       chainPriorityRpcs: "id",
       chainBackoffInterval: "id",
     })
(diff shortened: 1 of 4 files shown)


packages/connection-meta/src/TalismanConnectionMetaDatabase.ts
```diff
@@ -1,31 +1,31 @@
...
     // https://dexie.org/docs/Tutorial/Design#database-versioning
-    this.version(1).stores({
+    this.version(2).stores({
       // You only need to specify properties that you wish to index.
       // The object store will allow any properties on your stored objects but you can only query them by indexed properties
       // https://dexie.org/docs/API-Reference#declare-database
       //
       // Never index properties containing images, movies or large (huge) strings. Store them in IndexedDB, yes! but just don’t index them!
       // https://dexie.org/docs/Version/Version.stores()#warning
-      chainPriorityRpc: "id",
```
alecdwm, line 28: Might be good to set `chainPriorityRpc: null` so that the existing table is deleted

0xKheops: good catch thanks

will update and take the opportunity to merge setBestRpc and setWorstRpc into a single method

chidg (approved):
LGTM

PR #1585, opened 2024-08-15
wip: inline staking

0xKheops:

PR #1600, opened 2024-09-05
fix: onramp buy tokens picker

chidg:
- Show all possible tokens to buy, regardless of balances or active state
- Do not show balances
- Do not sort by any special sorting method
- Activate token/network if an inactive token/network is chosen

Previous `TokenPicker` and related components have lots of specific logic for send funds - I attempted to refactor these into abstract components but found it much easier to just create these new components which are more simple. In future we might start again, using these components as base and then adding further fuctionality with higher order components for send funds.


0xKheops:
Spotted a pretty bad suspense when opening the account dropdown, but it is (at least partially) due to useChainByGenesisHash which is fixed as part of #1585 
Would be nice to double-check after that one is merged

Note for later in case it didnt entirely fix it, adding a call to this preload should work:
```ts
const preloadAtom = atom((get) => Promise.all([get(addressBookAtom), get(accountsMapAtom)]))
```

0xKheops (dismissed):
Might want to delete apps/extension/src/ui/domains/Asset/TokenAmountField.tsx

Other than that, LGTM!

0xKheops (approved):
LGTM!

PR #1603, opened 2024-09-09
feat: erc20 approval set max instead of infinite

0xKheops:
In ERC20 Approval sign tx form, modifies the edit value drawer to provide a Max button instead of an Infinite button.

If clicked, the approval value will be set to the account's current balance.

### DIFF
--- a/apps/extension/src/ui/domains/Sign/Ethereum/shared/SignParamAllowanceButton.tsx
+++ b/apps/extension/src/ui/domains/Sign/Ethereum/shared/SignParamAllowanceButton.tsx
@@ -62,7 +62,7 @@ const isValidAmount =
     }
   }
 
-const useErc20Balance = (account: EvmAddress, token: Erc20TokenInfo) => {
+const useFetchErc20Balance = (account: EvmAddress, token: Erc20TokenInfo) => {
   const client = usePublicClient(token.evmNetworkId)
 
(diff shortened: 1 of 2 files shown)


apps/extension/src/ui/domains/Sign/Ethereum/shared/SignParamAllowanceButton.tsx
```diff
@@ -59,13 +62,33 @@ const isValidAmount =
     }
   }
 
+const useErc20Balance = (account: EvmAddress, token: Erc20TokenInfo) => {
```
chidg: It would be more ideal if this query would look in the balances pool first to see if there's a balance for the token before making the RPC call, is that possible?
If not, might be best to rename the hook (eg `useFetchErc20Balance`) because devs could get confused and assume this uses a balance already held locally

0xKheops: Hook renamed.
Kept it simple with just a contract balance fetch, because we dont want the whole balance subscriptions to start on the approval popup.
If we were to use balance libraries (which would do the same query, but with more overhead), we'd still need this contract call here as fallback in case the token is not registered in chaindata.

chidg (approved):
LGTM

PR #1616, opened 2024-09-13
Remove posthog libraries, use rest api for analytics

chidg:


packages/extension-core/src/domains/analytics/store.ts
```diff
@@ -0,0 +1,142 @@
...
+const TALISMAN_PROPERTIES = {
+  appVersion: process.env.VERSION,
+  appBuild: process.env.BUILD,
+  testBuild: DEBUG || ["dev", "qa", "ci"].includes(process.env.BUILD as string),
+}
+
+class AnalyticsStore extends StorageProvider<AnalyticsData> {
+  // the isReady promise prevents events being added to the queue while the analytics are being sent
+  isReady: Promise<boolean> = new Promise(() => true)
+
```
0xKheops, line 35: This never resolves, so i couldnt get my tests to work

so i tested with this instead: 
```ts
isReady: Promise<boolean> = new Promise((resolve) => { resolve(true) })
```

0xKheops: after i did i had this error once: 
image
image

packages/extension-core/src/domains/analytics/store.ts
```diff
@@ -0,0 +1,142 @@
...
+    await this.isReady
+
+    const sendInner = async () => {
+      if (!process.env.POSTHOG_AUTH_TOKEN) {
+        log.warn("POSTHOG_AUTH_TOKEN is not set, not sending analytics")
+        return true
+      }
+
+      const { data } = await this.get()
```
0xKheops, line 105: When testing, data here is undefined eventhough i captured a lot of events up front, so it stops there and never sends anything to posthog

0xKheops (changes requested):
Couldn't get it to work, please see below

0xKheops (approved):
LGTM, seems to work great!

PR #1621, opened 2024-09-17
fix: prevent to sending funds to ledger generic accounts

konan-5:


0xKheops (changes requested):
I was able to send Crust funds to a Kusama Ledger account, by pasting its substrate generic address and hitting the enter key.

In this case, it would be nice to block the user at the review screen page, and ideally explain why.

apps/extension/src/ui/domains/SendFunds/SendFundsAccountPicker.tsx
```diff
@@ -29,6 +30,10 @@ export const SendFundsAccountPicker = () => {
         .filter((account) => {
           if (!token) return false
 
+          if (!chain?.hasCheckMetadataHash && account.ledgerApp) {
```
chidg, line 33: This check isn't quite right. `account.ledgerApp` can be `substrate-legacy` or `substrate-generic` (see `SubstrateLedgerAppType`. If the chain does not have `hasCheckMetadataHash === true`, then we should allow accounts with `ledgerApp === 'substrate-legacy'`, as long as the genesis hash of the account matches the genesis hash of the chain.

0xKheops:
I think that this has already been fixed as part of #1583

this commit: https://github.com/TalismanSociety/talisman/pull/1583/commits/33b609ce1a6ae911f345bdb6157a539f2aae1b44

PR #1628, opened 2024-10-04
fix: evm addresses in custom xcm transfer sign ui

0xKheops:
Before:
image

After:
image


alecdwm (approved):
<div align="center">

LGTM

</div>

PR #1630, opened 2024-10-04
fix: ledger evm derivation path type dropdown

0xKheops:
Fixes the problem where changing the selected item seemed to have no effect (eventhough it did)

Took the opportunity to rename variables to make more sense


alecdwm (approved):
LGTM!

PR #1640, opened 2024-10-08
fix: tx history bugs

0xKheops:
- prevents speed-up/cancel options to be displayed on txs with a nonce lower than current
- opens explorer links in blank tab
- better cleanup of unknown transactions on startup


alecdwm (approved):
LGTM!

PR #1647, opened 2024-10-11
feat: new settings layout

0xKheops:
Made all pages pure to ensure any suspense is caught by the main content area, and sidebars don't flicker
ex: 
```tsx
export const AccountAddJsonPage = () => (
  <DashboardLayout sidebar="settings" width="660">
    <Content />
  </DashboardLayout>
)
```


chidg (approved):
LGTM

PR #1649, opened 2024-10-15
fix: timeout on asset discovery balance checks

0xKheops:


packages/extension-core/src/domains/assetDiscovery/scanner.ts
```diff
@@ -192,42 +194,20 @@ class AssetDiscoveryScanner {
...
+                    throw new Error(`No token found for tokenId ${check.tokenId}`)
                   }
 
-                  const balance = await client.getBalance({
-                    address: check.address as EvmAddress,
-                  })
-                  return balance.toString()
+                  return await Promise.race([
+                    getEvmTokenBalance(client, token, check.address as EvmAddress),
+                    throwAfter(5000, "timeout"),
```
chidg: Can we do
```
throwAfter(5000, `Asset Scan Timeout - ${check.tokenId}`),
``` 
instead? Since this will end up in Sentry

chidg (approved):
LGTM

PR #1661, opened 2024-10-16
fix: evm issues bundle

0xKheops:
- fixes balances for users with a lot of evm accounts
- fixes erc20 balances for customised evm networks
- leverages a contract to fetch batch of balances, in asset discovery


chidg (approved):
LGTM!

PR #1664, opened 2024-10-18
feat: yeet jotai

0xKheops:
Replaces `jotai` by `react-rxjs`, resulting in much less suspenses occuring and better performance

Also fixes bugs found along the way: 
- deleting an account could make the accounts page crash (based on which subscription fires first between accounts and accountsCatalog)
- portfolio tokens toolbar not appearing on first render


chidg (dismissed):
LGTM!

chidg (approved):
LGTMA!

PR #1672, opened 2024-10-22
chore: dependencies update 202410

0xKheops:
updated everything except : 
- eslint : too much work, i ran out of time
- jest-webextension-mock : breaks password related test, couldn't figure this out in time
- node in .npmrc : same, it breaks password related tests, couldn't find why

Also ran a `pnpm audit --fix` which added the following entries to pnpm overrides, in the root `package.json`

image


apps/extension/src/ui/domains/Account/ManageAccounts/TreeItems.tsx
```diff
@@ -108,3 +124,106 @@ const SeparatorDropZone: FC<{
     </TreeDroppable>
   )
 }
+
+export const TreeItemFolder: FC<{
```
0xKheops, line 128: moved this here to fix a circular dependency

apps/balances-demo/tailwind.config.cjs
```diff
@@ -5,8 +5,8 @@ const TALISMAN_TAILWIND_CONFIG = require("talisman-ui/tailwind.config.cjs")
 module.exports = {
   ...TALISMAN_TAILWIND_CONFIG,
   content: [
-    "./**/*.{html,ts,tsx,svg}",
+    "./src/**/*.{html,ts,tsx,svg}",
```
0xKheops, line 8: this was scanning node_modules

apps/extension/src/ui/apps/dashboard/routes/Settings/Mnemonics/MnemonicRenameModal.tsx
```diff
@@ -31,7 +31,7 @@ const MnemonicRenameForm: FC<{
     () =>
       yup
         .object({
-          name: yup.string().required("").notOneOf(otherAccountNames, t("Name already in use")),
+          name: yup.string().required(" ").notOneOf(otherAccountNames, t("Name already in use")),
```
0xKheops, line 34: something changed in yup.required()
we used to be able to not print an error message by forcing an empty string but this doesnt work anymore.
so i added a space instead, everywhere we dont want the 'FIELDNAME is a required field!' default error message

apps/extension/src/ui/apps/dashboard/routes/Tokens/AddCustomTokenPage.tsx
```diff
@@ -48,21 +42,24 @@ const Content = () => {
       yup
         .object({
           evmNetworkId: yup
-            .string()
-            .required()
-            .oneOf(
-              networks.map(({ id }) => id),
-              t("Invalid network")
-            ),
+            .mixed<EvmNetworkId>(
```
0xKheops, line 45: had to adjust most schemas because now the schema type and the form data type must match, when using react-hook-forms

apps/playground/tsconfig.node copy.json
```diff
@@ -0,0 +1,20 @@
+{
```
alecdwm: I think the name of this file is wrong?

packages/balances-react/package.json
```diff
@@ -36,28 +36,28 @@
...
-    "react": "^18.2.0",
-    "ts-jest": "^29.1.1",
-    "typescript": "^5.2.2"
+    "react": "^18.3.1",
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.3"
   },
   "peerDependencies": {
-    "@polkadot/util-crypto": "12.x",
+    "@polkadot/util-crypto": "13.2.1",
```
alecdwm: I think we want peerDependencies to remain as `13.x` so we use whatever version is installed on the dapp - rather than pulling in an additional copy of util-crypto just for this library

packages/balances/package.json
```diff
@@ -37,36 +37,36 @@
...
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.3"
   },
   "peerDependencies": {
-    "@polkadot/api-contract": "10.x",
-    "@polkadot/types": "10.x",
-    "@polkadot/util": "12.x",
-    "@polkadot/util-crypto": "12.x",
-    "@substrate/txwrapper-core": "4.x"
+    "@polkadot/api-contract": "14.1.1",
```
alecdwm: Same deal for peerDeps

packages/chain-connector/package.json
```diff
@@ -33,23 +33,23 @@
...
-    "typescript": "^5.2.2"
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.3"
   },
   "peerDependencies": {
-    "@polkadot/rpc-provider": "10.x",
-    "@polkadot/util": "12.x",
-    "@polkadot/x-global": "12.x",
-    "@polkadot/x-ws": "12.x"
+    "@polkadot/rpc-provider": "14.1.1",
```
alecdwm: peerDeps

packages/on-chain-id/package.json
```diff
@@ -31,25 +31,23 @@
...
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.3"
   },
   "peerDependencies": {
-    "@polkadot/api": "10.x",
-    "@polkadot/api-contract": "10.x",
-    "@polkadot/types": "10.x",
-    "@polkadot/rpc-provider": "10.x",
-    "@polkadot/util": "10.x"
+    "@polkadot/api-contract": "14.1.1",
```
alecdwm: peerDeps

packages/util/package.json
```diff
@@ -28,24 +28,24 @@
...
-    "ts-jest": "^29.1.1",
-    "typescript": "^5.2.2"
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.3"
   },
   "peerDependencies": {
-    "@polkadot/keyring": "12.x",
-    "@polkadot/util": "12.x",
-    "@polkadot/util-crypto": "12.x",
+    "@polkadot/keyring": "13.2.1",
```
alecdwm: peerDeps

packages/util/src/twox64Concat.ts
```diff
@@ -4,5 +4,6 @@ import { xxhashAsU8a } from "@polkadot/util-crypto"
 const bitLength = 64
 
 export function twox64Concat(input: string | Buffer | Uint8Array): `0x${string}` {
-  return u8aToHex(u8aConcat(xxhashAsU8a(input, bitLength), u8aToU8a(input)))
+  const inputAsU8a = typeof input === "string" ? input : new Uint8Array(input)
```
alecdwm, line 7: ~~I'm not sure about this - I suspect when `input` is a `string`, it might be a hex string?
In which case, I don't think `new Uint8Array(hexString)` correctly decodes a hex string into a Uint8Array.~~

Ah I read this the wrong way around - we're using `new Uint8Array` to turn a `Buffer` into a `Uint8Array`. Looks good!

apps/playground/tsconfig.node copy.json
0xKheops: fixed, thanks!

alecdwm (dismissed):
LGTM! 🔥

alecdwm (approved):

apps/extension/src/ui/apps/dashboard/routes/Settings/Mnemonics/MnemonicRenameModal.tsx
chidg: Maybe worth a try? https://github.com/jquense/yup?tab=readme-ov-file#error-message-customization

PR #1675, opened 2024-10-24
Remove livequery from tokenrates subscriptions

chidg:
Intended functionality:
- token rates are fetched from DB when wallet UI is first opened
- token rates continue to poll and update regularly while wallet UI is open, and are persisted to DB
- if token rates DB is cleared, token rates can be fetched and updated without error


packages/extension-core/src/domains/tokenRates/store.ts
```diff
@@ -16,12 +16,14 @@ import { activeTokensStore, filterActiveTokens } from "../tokens/store.activeTok
 const MIN_REFRESH_INTERVAL = 5 * 60_000
 
 // refresh token rates while sub is active every 10 minutes
-const REFRESH_INTERVAL = 10 * 60_000
+const REFRESH_INTERVAL = 10 * 3000 //60_000
```
alecdwm:
```suggestion
const REFRESH_INTERVAL = 30_000 // 30_000ms == 30 seconds
```

packages/extension-core/src/domains/tokenRates/store.ts
```diff
@@ -16,12 +16,14 @@ import { activeTokensStore, filterActiveTokens } from "../tokens/store.activeTok
 const MIN_REFRESH_INTERVAL = 5 * 60_000
 
 // refresh token rates while sub is active every 10 minutes
```
alecdwm:
```suggestion
// refresh token rates every 30 seconds while sub is active
```

0xKheops (approved):
LGTM!

PR #1676, opened 2024-10-25
chore: apply prettier config everywhere

0xKheops:
- result of `pnpm prettier --write .`
- also deleted useAlec.ts (unused, and prettier changes breaks eslint rules)

+ tsconfig fix for 3 packages


chidg (approved):
LGTM!

PR #1682, opened 2024-11-03
feat: TAO staking

UrbanWill:
# Description

Implements TAO staking

##  ⚠️ Warning 
- Depends on https://github.com/TalismanSociety/talisman-config/pull/7
- Add `TAOSTATS_API_KEY` to .env

## Techical 

- Renamed folder/components from Nomination pool bond specific to generic bonding names, eg: `NomPoolBondModal` -> `BondModal`
- Created bittensor and nompool folders inside a new hooks folder for organization purposes
- Created `useGetStakeInfo` and `useGetUnbondInfo` hooks to support the wizard hooks. These hooks are responsible for returning all the information that is specific to an Asset that is being staked/unbonded.
- Removed fake fee estimate, use a fee estimation that defaults to `minJoinBond` if the input amount is 0 or less than min for that pool, disable form submission if `planks` is < 0n

## TODO's

- [x] Hooks for fetching TAO delegator data
- [x] Pop up UI for selecting delegator
- [x] Dashboard UI for selecting a delegator
- [x] Refactor dashboard asset view to display the amount stake per Delegator
- [x] Hook to track the block number when the user unstaked the total balance from a delegator (used for 360 blocks cooldown) 

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/ee0fac2a-90cb-4a4d-a395-866e4a26e21c


alecdwm (dismissed):
👌

chidg:
Looking great, a few small things to add re: error handling of fetch requests.

When no `TAOSTATS_API_KEY` is configured, or else the `fetchBittensorInfiniteValidators` function fails for some reason, it would be good to display a nice message like "Unable to fetch validators" rather than just an empty screen

Screenshot 2024-11-19 at 6 24 05 PM

apps/extension/src/ui/domains/Staking/hooks/bittensor/useGetBittensorSupportedDelegates.ts
```diff
@@ -0,0 +1,22 @@
+import { useQuery } from "@tanstack/react-query"
+
+import { ValidatorsResponse } from "./types"
+
+const SUPPORTED_DELEGATES_URL =
+  "https://raw.githubusercontent.com/opentensor/bittensor-delegates/main/public/delegates.json"
+
+const fetchBittensorSupportedDelegates = async (): Promise<ValidatorsResponse> => {
```
chidg, line 8: Please handle failures of this request

apps/extension/src/ui/domains/Staking/hooks/bittensor/useGetBittensorValidator.ts
```diff
@@ -0,0 +1,108 @@
+import { useQueries, useQuery } from "@tanstack/react-query"
+import { TAOSTATS_API_KEY, TAOSTATS_BASE_PATH } from "extension-shared"
+
+const fetchBittensorValidator = async (hotkey: string | undefined): Promise<BittensorValidator> => {
+  try {
+    return await (
+      await fetch(`${TAOSTATS_BASE_PATH}/api/validator/latest/v1?hotkey=${hotkey}`, {
```
chidg, line 17: This one does handle errors internally, but the error state doesn't seem like it's properly handled in the calling logic - see comment on `BondForm.tsx`

apps/extension/src/ui/domains/Staking/hooks/bittensor/useGetBittensorInfiniteValidators.ts
```diff
@@ -0,0 +1,33 @@
+import { useInfiniteQuery } from "@tanstack/react-query"
+import { TAOSTATS_API_KEY, TAOSTATS_BASE_PATH } from "extension-shared"
+
+import { ValidatorsData } from "./types"
+
+const MAX_PAGE_SIZE = 100
+
+const fetchBittensorInfiniteValidators = async (page: number = 1): Promise<ValidatorsData> => {
+  return await (
+    await fetch(
```
chidg: Please handle errors in this request, or if the api key is not set

apps/extension/src/ui/domains/Staking/hooks/bittensor/useCombineBittensorStakeInfo.ts
```diff
@@ -0,0 +1,123 @@
...
+  balances: Balances
+}
+
+export const useCombineBittensorStakeInfo = ({ address, balances }: CombineBittensorStakeInfo) => {
+  const tokenId = "bittensor-substrate-native"
+  const token = useToken(tokenId)
+  const tokenRates = useTokenRates(tokenId)
+  const selectedCurrency = useSelectedCurrency()
+
+  const totalStaked = balances.each.reduce((acc, b) => {
```
0xKheops: missing useMemo (checked with a console log and this hook is called a lot, even on non TAO token screens)
same with lines 29 and 42

apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
```diff
@@ -24,26 +24,36 @@ import { Fiat } from "../../Asset/Fiat"
...
+import { BondAccountPicker } from "./BondAccountPicker"
+import { BondAccountPillButton } from "./BondAccountPillButton"
+import { useBondWizard } from "./useBondWizard"
 
 const AssetPill: FC<{ token: Token | null }> = ({ token }) => {
   const { t } = useTranslation()
 
   if (!token) return null
 
+  const stakeAssetLabel = () => {
```
0xKheops, line 41: these labels need to be wrapped in t("") from here, or our localization parser wont pick them up

apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
```diff
@@ -275,16 +286,56 @@ export const AmountEdit = () => {
             <div className="text-brand-orange line-clamp-2 text-center text-xs">
               {inputErrorMessage}
             </div>
+            {stakeWarningMessage && (
+              <div>
+                <Tooltip>
+                  <TooltipTrigger asChild>
+                    <div className="text-alert-warn flex items-center justify-center gap-1 whitespace-nowrap text-xs">
+                      <InfoIcon className="inline-block" /> {t(stakeWarningMessage)}
```
0xKheops: t() should always contain the string itself, can't be variable or parser won't pick it up

apps/extension/src/ui/domains/Staking/Bond/useBondButton.ts
```diff
@@ -20,14 +20,28 @@ export const useNomPoolBondButton = ({
   const ownedAccounts = useAccounts("owned")
   const token = useToken(tokenId)
   const remoteConfig = useRemoteConfig()
-  const { open } = useNomPoolBondModal()
+  const { open } = useBondModal()
+
+  // TODO: Add this to talisman-config
+  const mockedRemoteConfig = {
```
0xKheops: remove ?

apps/extension/src/ui/domains/Staking/Bond/useBondWizard.ts
```diff
@@ -313,10 +250,11 @@ export const useNomPoolBondWizard = () => {
       })
 
     return null
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
0xKheops: any reason for this ?

apps/extension/src/ui/domains/Staking/hooks/bittensor/useBittensorStakedBlockNumber.ts
```diff
@@ -0,0 +1,48 @@
...
+type AccountStakeData = {
+  [delegator: string | number]: number
+}
+
+type StoredData = {
+  [account: string]: AccountStakeData
+}
+
+export const useBittensorStakedBlockNumber = () => {
+  const [storedValue, setStoredValue] = useLocalStorage<StoredData | null>("initialValue", null)
```
0xKheops: Please provide a better name to this store, we need to be able to tell what it's for by it's name, when looking at all the storages entry of the domain.
Also not sure about the fact that we should use local storage like this at all, we usually implement our stores in extension-core with `StorageProvider` or one of it's derivatives.
@chidg wdyt ?

apps/extension/src/ui/domains/Staking/hooks/bittensor/useBittensorStakedBlockNumber.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // Method to get data by account and delegator
+  const getByAccountAndDelegator = ({
```
0xKheops: missing useCallback
same below

apps/extension/src/ui/domains/Staking/hooks/bittensor/useCanStakeBittensor.ts
```diff
@@ -0,0 +1,52 @@
...
+    useGetBittensorTotalHotkeyColdkeyStakes({
+      isEnabled: chainId === "bittensor",
+      sapi,
+      address: address,
+      hotkey: hotkey,
+    })
+
+  const { getByAccountAndDelegator } = useBittensorStakedBlockNumber()
+
+  const unbondBlock = getByAccountAndDelegator({ account: address, delegator: hotkey })
```
0xKheops: missing useMemo

apps/extension/src/ui/domains/Staking/hooks/bittensor/useGetBittensorValidator.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export const useGetBittensorValidators = ({
+  hotkeys,
+  isEnabled = true,
+}: {
+  hotkeys: (string | undefined)[]
+  isEnabled?: boolean
+}) => {
+  return useQueries({
```
0xKheops, line 46: That's very sexy ser :)

apps/extension/src/ui/domains/Staking/shared/BondPoolName.tsx
```diff
@@ -0,0 +1,71 @@
...
+}) => {
+  const { setStep, step } = useBondWizard()
+
+  let data,
+    isLoading = false,
+    isError = false,
+    poolName,
+    defaultPoolName = "Talisman Pool"
+
+  const hookMap = {
```
0xKheops: Please refactor with 2 child compoenents, selectively rendered depending on chainId
this will remove the need for the hook map and "let" variables in component body, for easier maintenance and readability
hook calls should never be conditional.

apps/extension/src/ui/domains/Staking/shared/StakingFeeEstimate.tsx
```diff
@@ -18,7 +18,7 @@ export const StakingFeeEstimate: FC<{
         <div className={classNames("text-alert-error truncate", className)}>
           Failed to estimate fee
         </div>
-      ) : plancks && tokenId ? (
+      ) : (plancks || plancks === 0n) && tokenId ? (
```
0xKheops, line 21: a 0 fee would be a bug, how did this happen ?

apps/extension/src/ui/domains/Staking/shared/StakingUnbondingPeriod.tsx
```diff
@@ -1,48 +1,36 @@
...
-            locale,
-          })
-        : t("N/A"),
-    [duration, locale, t],
-  )
+  let data,
+    isLoading = false,
+    duration = t("N/A")
+
+  const hookMap = {
```
0xKheops: same as BondPoolName

apps/extension/src/ui/domains/Transactions/TxProgress.tsx
```diff
@@ -118,6 +118,7 @@ const useTxStatusDetails = (tx?: WalletTransaction) => {
           animStatus: "failure",
         }
       case "success":
+        onSuccess?.(tx.blockNumber)
```
0xKheops: please move in it's own useEffect based on the single tx.status dependency, or this will be called multiple times
also i think this useEffect should be in TxProgressBase, as it's unrelated with tx details to be displayed

packages/util/src/formatTokenDecimals.ts
```diff
@@ -0,0 +1,4 @@
+export const formatTokenDecimals = (amount: number | string, decimals: number): number => {
```
0xKheops: please remove and use plancksToTokens instead
this one seems unsafe, notably it looks like it could output more decimals than the token supports

0xKheops (changes requested):
some minor things, please check below.

another that didnt fit to a particular part of the code is that i see bittensor staking related hooks being called a lot for no reason, especially in AssetDetails
calling any of them loads bittensor sapi object which downloads the full metadata.
Unless a bittensor balance is detected, it would be nice to make sure those hooks do nothing. 
Note: at this time even if if bittensor is disabled in settings, they will fetch metadata and execute all the logic

apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
```diff
@@ -275,16 +286,56 @@ export const AmountEdit = () => {
...
+    apr = 0
+
+  const hookMap = {
+    nominationPool: useStakingAPR,
+    bittensor: useGetBittensorValidator,
+  }
+
+  switch (token?.chain?.id) {
+    case "bittensor":
+      ;({ data, isLoading } = hookMap["bittensor"](poolId as unknown as string))
```
chidg: Not such a big fan of this syntax for this, but either way would prefer to include the `isError`state here so we can show an appropriate message

apps/extension/src/ui/domains/Staking/shared/BondPoolName.tsx
```diff
@@ -0,0 +1,71 @@
...
+    nominationPool: useNomPoolName,
+    bittensor: useGetBittensorValidator,
+  }
+
+  switch (chainId) {
+    case "bittensor":
+      ;({ data, isLoading, isError } = hookMap["bittensor"](poolId as unknown as string))
+      poolName = data?.data?.[0].name || ""
+      poolName = (
```
chidg: Duplicate variable assignment, is that intentional?

apps/extension/src/ui/domains/Staking/shared/useGetLatestBlockNumber.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ScaleApi } from "@ui/util/scaleApi"
+
+type GetLatestBlockNumber = {
+  sapi: ScaleApi | undefined | null
+  isEnabled: boolean
+}
+
+export const useGetLatestBlockNumber = ({ sapi, isEnabled }: GetLatestBlockNumber) => {
+  return useQuery({
+    queryKey: ["useGetLatestBlockNumber", sapi?.id],
```
chidg, line 12: I got an error on this line when testing (no Taostats API key configured) - `Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["useGetLatestBlockNumber",null]`

chidg (changes requested):
Few more comments, apologies - these got lost in vs code PR review process somewhere

apps/extension/src/ui/domains/Staking/hooks/bittensor/useGetBittensorValidator.ts
UrbanWill: Thanks!
Edit: no, you are! ❤️

apps/extension/src/ui/domains/Staking/shared/StakingFeeEstimate.tsx
UrbanWill: That's actually the extrinsic fee for unstaking TAO, check this tx as an example https://taostats.io/extrinsic/4264248-0007?network=finney

apps/extension/src/ui/domains/Staking/Bond/useBondButton.ts
UrbanWill: Can remove when this gets an ✅  and merged: https://github.com/TalismanSociety/talisman-config/pull/7

apps/extension/src/ui/domains/Transactions/TxProgress.tsx
```diff
@@ -241,14 +256,21 @@ const TxProgressEvm: FC<TxProgressEvmProps> = ({ tx, className, onClose }) => {
...
+  onSuccess?: (...args: T) => void
 }
 
-export const TxProgress: FC<TxProgressProps> = ({ hash, networkIdOrHash, onClose, className }) => {
+export const TxProgress = <T extends unknown[]>({
+  hash,
+  networkIdOrHash,
+  onClose,
+  className,
+  onSuccess,
```
0xKheops: Sorry i didnt dig this thing enough last time. 
It feels like success detection shouldnt originate from here, which is a UI component that is meant to support both EVM and substrate.

Option 1 : use useTransaction(hash) from useUnbondWizard() to detect success from there, without needing any callbacks, and rollback all changes to this component
Option 2 : move the success check from TxProgressSubstrate up to TxProgress, so it can support both substrate and evm

Preference for option 1 though

0xKheops (dismissed):
Sorry, one more change needed in TxProgress :)

apps/extension/src/ui/domains/Staking/hooks/bittensor/useBittensorStakedBlockNumber.ts
chidg: ah yeah, should have spotted this, generally we should not use the normal window local storage api in the extension. Better to add a key to one of the existing stores (probably `app` in this case)

apps/extension/src/ui/domains/Transactions/TxProgress.tsx
UrbanWill: Went with option um, thanks for the suggestion!

apps/extension/src/ui/domains/Staking/hooks/bittensor/useCanStakeBittensor.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const [, lastStakedBlockNumber] = useMemo(() => stakeData ?? [0n, 0n], [stakeData])
+
+  useEffect(() => {
+    if (chainId !== "bittensor") return
+    const latestActionBlock = Math.max(
+      Number(unbondBlockNumber ?? 0),
+      Number(lastStakedBlockNumber ?? 0),
+    )
+    const coolDownBlockNumber = Number(latestActionBlock) + STAKE_INTERVAL_BLOCKS
```
chidg, line 48:
```suggestion
    const coolDownBlockNumber = latestActionBlock
      ? Number(latestActionBlock) + STAKE_INTERVAL_BLOCKS
      : 0
```

chidg: This prevents a flash of the `canStake === false` state when the screen first loads

chidg (dismissed):
LGTM!

chidg (approved):
LGGGGTM

PR #1686, opened 2024-11-06
fix: connected accounts issues

0xKheops:
- Fixes an invalid count of connected accounts, due to deleted accounts still being listed in site authorizations store
- Optimistic updates when changing connected accounts


alecdwm (approved):
<div align="center">

LGTM

</div>

PR #1688, opened 2024-11-08
feat/allow-other-substrate-block-explorers

0xKheops:


chidg (approved):
LGTM
Comments/questions:
- should we have an info text in the form to explain that the url schema must match Subscan?
- should we/how can we update the `subscanUrl` property in chaindata without causing problems?

0xKheops:
> LGTM Comments/questions:
> 
> * should we have an info text in the form to explain that the url schema must match Subscan?
> * should we/how can we update the `subscanUrl` property in chaindata without causing problems?

Thanks!

yes we could add a description, though i think it's not a user problem. It should be obvious that all substrate explorers should align with subscan url schemes to be compatible with existing wallets.
I think changing the property name would be risky. if we go ahead and do it, it would be nice to go with a `explorerUrls` property of type Array<String>, to be future proof.

PR #1694, opened 2024-11-14
feat: asset discovery improvements

0xKheops:
- Automatically enable discovered assets
- Account context menus now include a : "Scan missing tokens" button that triggers a full scan for that account


chidg (approved):
LGTM

PR #1695, opened 2024-11-15
feat: disable analytics on FF

0xKheops:
- Reverts changes to Chrome from #1692 
- Disables analytics and error reporting on FF
- Bonus: Hides Risk Analysis settings based on feature flag, as this was forgotten in #1677


chidg (approved):
LGTM

PR #1699, opened 2024-11-15
fix: improved autolock to only keep wallet unlocked while user is actively interacting with it

alecdwm:
Users are surprised when their wallet never autolocks - but the current behaviour intentionally only autolocks the wallet when all wallet UI has been closed.

This PR changes it that so the the wallet will autolock even if the UI is open (e.g. in a background tab) - provided that the user is not actively interacting with the UI (as measured by keydown / keypress / keyup / mousedown / mouseup / mousemove / touchstart / touchend / touchmove events).

NOTE: The embedded popup retains the old behaviour: the wallet will never lock for as long as the embedded popup is open.


chidg (approved):
LGTM

PR #1705, opened 2024-11-21
feat: parameterise orml tokens palletId

alecdwm:
Enables us to support orml tokens on Centrifuge.
For most chains, orml token balances can be retrieved via `Tokens.Account`, but for Centrifuge we must query `OrmlTokens.Account` instead.


chidg (commented):
LGTM

chidg (approved):

PR #1708, opened 2024-11-25
fix: decoding-adjustments

0xKheops:


chidg (approved):
Did these ever work?

LGTM!

PR #1714, opened 2024-11-26
Refactor/yeet-chaindataprovider-from-frontend

chidg:
Part of ongoing de-livequery-fication
Removed chaindataProvider from frontend, replacing livequeries with backend message subscriptions. ChaindataProvider should now only be instantiated in the backend and all chain data should be accessed via messages.


0xKheops (approved):
Looks and works great! LGTM!

PR #1719, opened 2024-11-29
fix: minor ui adjustments bundle

0xKheops:
- modal dialog header
- buy tokens dropdown options alignment
- nft tiles size adjustments to fit 4 per row (due to new dashboard layout width)


chidg (approved):
LGTM

PR #1723, opened 2024-11-29
fix: pre-release ui adjustments

0xKheops:
- removes the progress indicator on AD notification
- fixes the drawer parent in copy address modal


chidg (approved):
LGTM

PR #1724, opened 2024-12-02
fix: review builds via docker

alecdwm:
While we're waiting on https://github.com/polkadot-api/polkadot-api/pull/851 to make papi compatible with our monorepo *without* needing any hacks or workarounds, this PR fixes the existing workaround to also work in our docker-based review builds.

The existing workaround depends on `git` to revert the unwanted `pnpm-lock.yaml` changes which are made by `pnpm papi`.  
This PR introduces an alternative `pnpm papi:dockerbuildcompat` command which will instead make a copy of `pnpm-lock.yaml` before modification, then move the copy back over the modified file after papi has done its thing.


chidg (approved):
LGTM

PR #1725, opened 2024-12-02
fix: ledger add account loading balance state

0xKheops:
fixed inconsistant "flashing" balances in ledger add account flow
this will still happen though if mini metadatas update while the user is on that screen, as it restarts balance subs.


chidg (approved):
LGTM

PR #1727, opened 2024-12-02
feat: auto asset discovery

0xKheops:
- queue for scans
- removes all asset discovery related notifications
- after user enables one or more networks, automatically scans them
- after user adds one or more accounts, automatically scans them
- migration to clear current asset directory data stores upon upgrade
- moved the autoEnable.ts's logic inside the scanner so it can be executed between queued scans


chidg (approved):
LGTM!

PR #1729, opened 2024-12-05
fix: tao staking release ceremony bugs

UrbanWill:
# Description

- Disable useNomPoolStatus query if chain id is bittensor
- Do no throw if `TAOSTATS_API_KEY` is missing. This key should be used in dev mode only and not affect prod


chidg (approved):
LGTM

PR #1733, opened 2024-12-09
fix: quick fixes

alecdwm:
1. Adds a `data-amount` attribute for easier dev access to the exact token balance

<div align="center">

Screenshot 2024-12-09 at 13 38 16

</div>

---

2. Fixes the network infos rows to use the relay chain name (which we set) instead of the chainName (which is queried from the chain, and can be undesirable)

<div align="center">

| before | after |
| - | - |
|  |  |

</div>


0xKheops (approved):
LGTM!

PR #1736, opened 2024-12-10
fix: no sentry captures for scans and tx decoding

0xKheops:
- Use a vanilla ErrorBoundary for tx decoding
- Don't send "Failed to scan" errors to sentry
- Throttle reads to assetDiscovery store's observable


chidg (approved):
LGTM!

PR #1738, opened 2024-12-11
fix: more reliable autolock alarms

alecdwm:
Current behaviour:
- run autolock on initial delay of `x`, if missed for any reason then retry every `x` interval

New behaviour:
- run autolock on initial delay of `x`, if missed for any reason then retry every 30s


chidg (commented):
Concerned this could become a footgun in future, let's leave this one until we establish the real cause of reported problems

PR #1744, opened 2024-12-11
fix: dont use sapi for displaying balances

0xKheops:
- Changed native balance module to include one value per validator, in case the account is staking on multiple validators.
- Adjusted portfolio components to read hotkey from the balance rows instead of fetching them from chain


UrbanWill (dismissed):
Amazing! ❤️

chidg (approved):
LGTM!

alecdwm (approved):
LGTM!

PR #1746, opened 2024-12-12
feat: portfolio performance improvements

0xKheops:
- improves allBalances$ observable (one less pipe stage, and added a shareReplay as it s often read)
- removed the shareReplay on `authorisedSites$` observable, as it s wrapped in a bind (has a built-in shareReplay)
- optimised few useMemo/useCallback that had a dependency on `balances.each` (which changes on each read, nullifying the memo effect) to have a dependency on `balances` instead. 
- virtualized both popup and dashboard asset tables using `@tanstack/react-virtual`
- removed countUps on virtualized rows, as the count up animation would trigger too often
- virtualized rows in token picker
- virtualized rows in tx history


UrbanWill (dismissed):
LGTM!

However I am kinda sad seeing the count up go, I've create an example where we can keep it while virtualizing the rows without too much change in the current virtualization.

Have a look and let us know if you like it: https://github.com/TalismanSociety/talisman/pull/1750

UrbanWill (approved):
🔥

PR #1749, opened 2024-12-16
feat: portfolio header adjustments

0xKheops:
- Disable swap button for watched accounts
- Receive button labelled as “Copy” if current account is a watched account + enable it


chidg (approved):
LGTM

PR #1750, opened 2024-12-16
Refactor/keep count up perf improvements 202412

UrbanWill:
# Description

TLDR; same UX but now virtualized

I think our wallet looks a bit too pedestrian without the count up when first loading assets. 

I created this example where we can keep the virtualized rows and the count up, which gets disabled when the user scrolls past a our estimate size * 3. 

## Example 1: Popup view
https://github.com/user-attachments/assets/07e0221d-11e5-4b89-841b-8079cf351181

## Example 2: Dashboard asset row Fiat values 
https://github.com/user-attachments/assets/31b1decc-d852-4290-981c-99e6e935d996


chidg:
Closing because similar fixes included in #1746

PR #1751, opened 2024-12-16
fix: staking buttons adjustments

0xKheops:
- don't show staking button for accounts that are solo-staking
- don't show unbond button on crowdloan rows
- don't render staking button inside balance rows


chidg (approved):
LGTM

PR #1757, opened 2024-12-17
fix: tx decoding - display non-ascii values as hex

0xKheops:
fixes a TODO that i forgot to process when implementing advanced tx decoding : 
when displaying the payload as json/yaml, we dont know what binary objects contain.
if the binary.toText() doesnt return ascii text, it will now display the hex output instead.

before : 
Screenshot 2024-12-17 at 2 14 21 PM

after : 
Screenshot 2024-12-17 at 2 13 53 PM


chidg (approved):
LGTM

PR #1760, opened 2024-12-19
feat: add account redesign

alecdwm:
Implements new designs for `/accounts/add`, including a network picker for users who don't know whether they need to create an Ethereum or a Polkadot account for use with a given chain.


UrbanWill (approved):
LGMT,

However there is this weird yo-yo effect when scrolling the virtualized dropdown networks list too fast. I tried to fix it myself but seems like react-virtual lib is not playing nice with headlessUi components. Time to give shadcn a go?

Don't think it is a blocker though, since the list is quite big and users will make use of the search functionality.

In the following video I never scrolled up:


https://github.com/user-attachments/assets/384ddcf0-e59e-4949-ae38-f4ff5322dbf7

PR #1765, opened 2024-12-24
fix: use tokenId instead of symbol for send funds lookup where possible

alecdwm:
Prevents a bug when the same symbol is used by multiple tokens on one network.

e.g. `USDC` vs `USDC.e` vs `whUSDC` (which is also just `USDC` in the contract) on Polygon.

### DIFF
--- a/apps/extension/src/ui/domains/Portfolio/AssetDetails/SendFundsIconButton.tsx
+++ b/apps/extension/src/ui/domains/Portfolio/AssetDetails/SendFundsIconButton.tsx
@@ -1,42 +1,25 @@
-import { ChainId, EvmNetworkId } from "@talismn/chaindata-provider"
+import { TokenId } from "@talismn/chaindata-provider"
 import { SendIcon } from "@talismn/icons"
 import { useCallback } from "react"
 import { Tooltip, TooltipContent, TooltipTrigger } from "talisman-ui"
 
 import { useSendFundsPopup } from "@ui/hooks/useSendFundsPopup"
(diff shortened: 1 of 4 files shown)


apps/extension/src/ui/domains/Portfolio/AssetDetails/SendFundsIconButton.tsx
```diff
@@ -10,23 +10,32 @@ import { isTransferableToken } from "@ui/util/isTransferableToken"
...
+          isTransferableToken(t) &&
+          (("evmNetwork" in t && t.evmNetwork?.id === networkId) || t.chain?.id === networkId)
+      : // search by token symbol
+        (t) =>
+          t.symbol === symbol &&
+          isTransferableToken(t) &&
+          (("evmNetwork" in t && t.evmNetwork?.id === networkId) || t.chain?.id === networkId),
   )
 
   const { canSendFunds, cannotSendFundsReason, openSendFundsPopup } = useSendFundsPopup(
```
chidg, line 24: `useSendFundsPopup` takes `tokenSymbol` as a prop, this suggests two things:
1. We should stick with the same naming convention for props (so call it `tokenSymbol` instead of `symbol` in the props for this hook)
2. rather than performing a `find` if the token symbol is given, and then giving the first token found to the hook, we should just pass the token symbol to the hook directly, and allow the logic in the hook to determine which token/s to use (actually it's the logic in the backend, as the symbol is just passed through)

chidg (approved):
LGTM

PR #1767, opened 2024-12-24
fix: weird scroll behaviour in AccountTypeNetworkSearch

alecdwm:
I originally used `@tanstack/react-virtual` instead of the built-in virtual prop of `<Combobox />` because I needed the `Network | Account Type` header inside of the dropdown list.

This PR implements an alternative solution; it uses the built-in virtual prop instead of `@tanstack/react-virtual`, and then implements the header as a faux `<ComboboxOption />`.

As far as I can tell this fixes the weird scrolling behaviour.


UrbanWill (dismissed):
LGTM! 🔥 

Pushed to minor fixes:

1- Dropdown had a fixed size, when the search result has fewer options the dropdown should be smaller. 

Screen shot before the fix:
Screenshot 2024-12-30 at 10 13 08

2- Searching and then hitting enter caused the app to crash because an empty option is still  a default combo box option. Created a handleChange fn for that case, and tidied up side effects inside this fn as well.

Example before the fix: 

https://github.com/user-attachments/assets/e7c0ad1a-a8d1-4206-8617-7b2745f79cd9

chidg (approved):
LGTM!

PR #1773, opened 2025-01-03
fix: fetch metadata for immortal payloads

0xKheops:
Fixes sign popup crash on https://test.common.fi

### DIFF
--- a/packages/extension-core/src/util/getMetadataDef.ts
+++ b/packages/extension-core/src/util/getMetadataDef.ts
@@ -55,6 +55,10 @@ const getMetadataDefInner = async (
 ): Promise<TalismanMetadataDef | undefined> => {
   const [chain, genesisHash] = await getChainAndGenesisHashFromIdOrHash(chainIdOrHash)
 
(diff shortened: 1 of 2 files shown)


packages/extension-core/src/util/getMetadataDef.ts
```diff
@@ -55,6 +55,10 @@ const getMetadataDefInner = async (
 ): Promise<TalismanMetadataDef | undefined> => {
   const [chain, genesisHash] = await getChainAndGenesisHashFromIdOrHash(chainIdOrHash)
 
+  // blockHash being equal to genesisHash happens when trying to decode payload of immortal transactions
+  // in that case, blockHash must be used as block reference when fetching data
```
chidg: Comment doesn't quite make sense - if `blockHash` must be used as a reference, why is it being set to undefined?

0xKheops: good catch, thanks :)

chidg (approved):
LGTM

PR #1774, opened 2025-01-03
feat: new logo and `Error ID` in error boundary

alecdwm:
- [x] Added white title
- [x] Removed full-stops `.` from error boundary text
- [x] Updated error boundary logo
- [x] Added sentry error ID to error boundary, when available

| Old | New |
| --- | --- |
| old | new |

---

NOTE: We have the `<ErrorBoundary>` in three places, but I'm pretty sure that the top-level one is always present.  
Idk if we still need the lower-level ones @0xKheops?
Here's some screenshots showing which error boundaries are active on which pages of the UI:

Screenshot 2025-01-03 at 18 42 30
Screenshot 2025-01-03 at 18 42 37
Screenshot 2025-01-03 at 18 42 52
Screenshot 2025-01-03 at 18 43 03
Screenshot 2025-01-03 at 18 43 07
Screenshot 2025-01-03 at 18 43 14


0xKheops (approved):
LGTM !

About the 2 lower level ErrorBoundary i think they can be removed, would be nice to have this done as a separate PR though, to allow us to merge it when we have a coupe days time window to sustain test it.

PR #1775, opened 2025-01-06
feat: ledger enhancements

0xKheops:
- Fallbacks to webhid transport if fails to connect with webusb
- Display the custom Talisman error message, if any, instead of the generic "Failed to connect"
- Display the actual error message, if the error is unhandled.

image


chidg (approved):
LGTM!

PR #1776, opened 2025-01-06
fix: yeet redundant error boundaries

alecdwm:
Following on from https://github.com/TalismanSociety/talisman/pull/1774, this PR yeets the two extra uses of `<TalismanErrorBoundary>`, which we don't think we need anymore.


0xKheops (approved):
LGTYeet!

PR #1786, opened 2025-01-21
feat: export all accounts as json

0xKheops:
also includes a change to dashboard layout to:
- add `overflow-hidden` to main area
- replace `focus:visible` effects on most input fields to use borders instead of rings, to prevent them from being truncated when displayed in a `overflow-hidden` container


chidg (approved):
LGTM

PR #1788, opened 2025-01-21
fix: errors in balance pool

0xKheops:
Fixes a bunch of sentry errors about isActiveToken/isActiveEvmNetwork/isActiveChain being sometimes called from pool.ts with a null argument, that i can't reproduce on my computer. 

Also includes a fix to an unawaited promise in chaindata providers, that could be tied to the above problem


alecdwm (approved):
LGTM!

PR #1789, opened 2025-01-22
fix: normalize failed to update metadata error message

0xKheops:
simplifies the error message for when backend can't update metadata, so that:
- all errors have the same message and will be grouped together in sentry
- the chainId of target chain, if known, is added as extra property in the sentry payload


chidg (approved):
LGTM

PR #1794, opened 2025-02-06
chore: add rootnet uid to add/remove stake

UrbanWill:
## Description

Added rootnet uid to Bittensor add/remove stake, this change is required to comply with upcoming subnet staking.

Passing a netuid does not throw when using papi and is ignored by their network.

The following tx was submited with netuid, though it is not included in the params: 
https://taostats.io/extrinsic/4865692-0010?network=finney


chidg (approved):
LGTM

PR #1802, opened 2025-02-07
fix: don't throw error when currentEra is greater than 0, less than `maxErasToCheck`

alecdwm:
The code here would check e.g. eras `[-10, -9, -8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5]` if `currentEra` is 6.

Attempting to make calls to the chain with an `era` parameter less than `0` will throw errors.

Now, we'll filter for negative values, making the list of eras to check just `[0, 1, 2, 3, 4, 5]` if `currentEra` is 6.


0xKheops (approved):
LGTM!

PR #1818, opened 2025-03-05
fix: subnet name undefined

UrbanWill:
# Description

- Check if subnet name and id exists before displaying it
- Fixed N validators api calls. This was contributing to rate limiting issues


0xKheops (approved):
LGTM

PR #1824, opened 2025-03-12
fix: tao balances

UrbanWill:
# Description

## Fixes:
- Staked positions would wait 30s before fetching for the first time
- No position would be displayed if decoding a single dynamic info fails

## Feats
- Added retries
- Display staked position even if decoding the dynamic data for a subnet fails. In that case `alphaStakedInTao` cannot be calculated, so the value of `1n` is assigned so it can be displayed on the UI

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/047fdd72-03d7-480c-aec8-40b43c685399

### DIFF
--- a/.changeset/rotten-toys-accept.md
+++ b/.changeset/rotten-toys-accept.md
@@ -1,0 +1,5 @@
+---
+"@talismn/balances": patch
+---
+
+Fixed dTao staked balances fetching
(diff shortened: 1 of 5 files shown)


0xKheops (approved):
LGTM!

PR #1829, opened 2025-03-14
Feat/dtao balances UI

UrbanWill:
# Description

This PR implements displaying dTao balances as parachains.

## Technical

- Moved multiple components from `DashboardAssetDetails` and `PopupAssetDetails` to their own files, to make them easily reusable. The way God intended.
- Created `BittensorTokenBalances` components to handle displaying you guessed it, BIttensor chain token balances.
- Calculated `alphaToTaoRate` based on a single alpha token unit per Subnet and added to `subscribeSubtensorBalances`.
- Added toastats fetch hooks, fed initial data and created a custom hook to combine this data. Data combination will only occur if API calls are successful, else display initial data so users can still interact with staking/unstaking. This will become more relevant when dTao staking is implemented in the wallet.

#### Popup:

https://github.com/user-attachments/assets/4a81e208-9441-424d-adb2-5ee192a656a7

#### Dashboard:

https://github.com/user-attachments/assets/a92ccec3-077a-4332-b452-5be8e971373e


apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/BittensorTokenBalances.tsx
```diff
@@ -0,0 +1,114 @@
...
+    balances,
+  })
+
+  // wait for data to load
+  if (!chainOrNetwork || !summary || !token || balances.count === 0) return null
+
+  const groupedStakes = Object.groupBy(detailRows, ({ meta }) => meta?.netuid ?? CHAIN_INFO)
+
+  const sortedGroupedStakes = sortGroupedStakes(groupedStakes, CHAIN_INFO)
+
```
0xKheops: missing useMemo ?

apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/TokenBalancesDetailRow.tsx
```diff
@@ -0,0 +1,95 @@
...
+  symbol,
+  tokenId,
+  tokenDecimals,
+}: TokenBalancesDetailRowProps) => {
+  const alphaBalanceInTao = new BigNumber(
+    planckToTokens(row.meta?.amountStaked, tokenDecimals) || "0",
+  )
+
+  const tokenBalance = alphaBalanceInTao.gt(0) ? alphaBalanceInTao : row.tokens
+
```
0xKheops: missing useMemo?

apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { chainOrNetwork, summary, token, detailRows, status, networkType } = useTokenBalances({
+    tokenId,
+    balances,
+  })
+
+  // wait for data to load
+  if (!chainOrNetwork || !summary || !token || balances.count === 0) return null
+
+  if (tokenId === BITTENSOR_TOKEN_ID) {
+    return <BittensorTokenBalances tokenId={tokenId} balances={balances} />
```
0xKheops, line 28: perfect :)

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/AssetPercentageChange.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const AssetPercentageChange = ({
+  priceChange,
+  isLoading,
+  isError,
+}: AssetPercentageChangeProps) => {
+  const num = parseFloat(priceChange || "0")
+  const truncatedValue = num.toFixed(1)
+  const formattedValue = num > 0 ? `+${truncatedValue}` : truncatedValue
+
```
0xKheops: missing useMemo

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/BittensorTokenBalances.tsx
```diff
@@ -0,0 +1,146 @@
...
+  })
+
+  // wait for data to load
+  if (!chainOrNetwork || !summary || !token || balances.count === 0) return null
+
+  const groupedStakes = Object.groupBy(detailRows, ({ meta }) => meta?.netuid ?? CHAIN_INFO)
+
+  const sortedGroupedStakes = sortGroupedStakes(groupedStakes, CHAIN_INFO)
+
+  return sortedGroupedStakes.map(([key, groupedStakesByNetuid], i) => {
```
0xKheops: feels like the callback of the map should be a component, so it could leverage useMemo for everything that requires computation, especially the reduce function

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/TokenBalancesDetailRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  symbol: string
+  tokenId?: TokenId // unsafe, there could be multiple aggregated here
+  tokenDecimals: number
+}) => {
+  const alphaBalanceInTao = new BigNumber(
+    planckToTokens(row.meta?.amountStaked, tokenDecimals) || "0",
+  )
+
+  const tokenBalance = alphaBalanceInTao.gt(0) ? alphaBalanceInTao : row.tokens
+
```
0xKheops: missing useMemo

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { TokenBalancesList } from "./TokenBalancesList"
+
+const BITTENSOR_TOKEN_ID = "bittensor-substrate-native"
+
+type TokenBalancesParams = {
+  balances: Balances
+  tokenId: TokenId
+}
+
+export const TokenBalances = ({ balances, tokenId }: TokenBalancesParams) => {
```
0xKheops: maybe rename the TokenBalances folder to DashboardTokenBalances, to be consistent with the PopupTokenBalances one ?

apps/extension/src/ui/domains/Staking/Bittensor/constants.ts
```diff
@@ -1 +1,28 @@
+import { type BalanceDetailRow } from "@ui/domains/Portfolio/AssetDetails/useTokenBalances"
+
 export const ROOT_NETUID = 0
+
+export const CHAIN_INFO = "chainInfo"
+export const DTAO_LOGO =
+  "https://raw.githubusercontent.com/TalismanSociety/talisman-web/8c5d78676702214c6983b04c9f15c15f34d34109/packages/icons/src/svgs/dtaoLogo.svg"
+
+export const sortGroupedStakes = (
```
0xKheops: maybe move this function in its own file, as this one is called `constants.ts`

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useCombinedSubnetData.ts
```diff
@@ -0,0 +1,78 @@
...
+      {} as Record<number, SubnetData>,
+    )
+
+    setSubnetData(combinedSubnetData)
+  }, [
+    subnetDescriptionsData,
+    subnetDescriptionsData?.pages,
+    subnetPoolsData,
+    subnetPoolsData?.pages,
+  ])
```
0xKheops, line 72: being nitpicky here but i think  the useState and useEffect could be replaced by a single useMemo, to save a render cycle

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useGetInfiniteSubnetDescriptions.ts
```diff
@@ -0,0 +1,706 @@
...
+    getNextPageParam: (lastPage) => lastPage.pagination.next_page ?? undefined,
+    getPreviousPageParam: (firstPage) => firstPage.pagination.prev_page ?? undefined,
+    initialPageParam: 1,
+    staleTime: 5 * 60 * 1000, // 5 mins
+    gcTime: 10 * 60 * 1000, // 10 mins
+    refetchOnReconnect: true,
+  })
+}
+
+export const initialData: SubnetApiDescriptionsResponse = {
```
0xKheops, line 39: this will be hard to maintain, and we probably wont. Do we really need this initial data ?

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useGetInfiniteSubnetPools.ts
```diff
@@ -0,0 +1,2062 @@
...
+    },
+    getNextPageParam: (lastPage) => lastPage.pagination.next_page ?? undefined,
+    getPreviousPageParam: (firstPage) => firstPage.pagination.prev_page ?? undefined,
+    staleTime: 5 * 60 * 1000, // 5 mins
+    gcTime: 10 * 60 * 1000, // 10 mins
+    refetchOnReconnect: true,
+  })
+}
+
+export const initialData: SubnetApiResponse = {
```
0xKheops, line 36: same here, maintenance issue

0xKheops (changes requested):
Great one!
Just found few optimization opportunities, portfolio's performance being critical

apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/BittensorTokenBalances.tsx
UrbanWill: Wouldn't the memory cost of useMemo/useCallback be more expensive than doing the recalculations? I don't expect users to have that many positions.

Unless of course the asset details components re-renders a lot, then I am happy to wrap those in useMemo for now

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/TokenBalancesDetailRow.tsx
UrbanWill: Wouldn't the cost of useMemo outweigh the cost of recalculation of this variable?

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenBalances/AssetPercentageChange.tsx
UrbanWill: Is it really necessary? This is not an expensive calculation

apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/TokenBalancesDetailRow.tsx
UrbanWill: Is it really necessary? This is not an expensive calculation

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useGetInfiniteSubnetDescriptions.ts
UrbanWill: Yes, when Subnet staking gets implemented soon we need initial data to be displayed in case Taostats api calls fails. This will allow users to still stake on subnets with some limited information. On the maintenance side, I hope we don't need to touch this unless new subnets are added.

Heads up that will need to do the same for validators as well, but there we are fetching data directly from opentensor's repo.

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useGetInfiniteSubnetPools.ts
UrbanWill: Same as above ☝️

apps/extension/src/ui/domains/Portfolio/AssetDetails/PopupTokenBalances/TokenBalancesDetailRow.tsx
0xKheops: yeah i m being nitpicky, though when it comes to portfolio, any performance optimization is good to have.
some users have over 60 accounts, and balances observable fires a lot in some cases (ex if user owns some stETH), so every little quick-win like this helps :)

apps/extension/src/ui/domains/Staking/hooks/bittensor/dTao/useCombinedSubnetData.ts
UrbanWill: Yeah I see where you are coming from, but no.

We are feeding initial data into react-query hooks, that means we always have data returned by those hooks on their first call. However that is nota always guaranteed, since we subjected to possible rate limits, api changes, etc and in that case the data returned by the react-query hooks would be null/undefined. 

So to circumvent that, we store good data into state and if there is a bad api response we don't update the state and use the isError to partially render data and let the user know that that info may be stale (when we get to subnet staking).

TLDR; useMemo would return null if there's no data returned from apis, we need to display some subnet data on the UI.

0xKheops (dismissed):
LGTM!
though it would be nice to add a script that regenerates initial data objects, or remove them, someday™

0xKheops (approved):
LGTM!

PR #1830, opened 2025-03-21
feat: keyring upgrade banner

0xKheops:


packages/extension-core/src/domains/app/store.session.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// call this only from background.js
+export const resetSessionOnUnlock = () => {
+  let isLocked = false
+
+  // first ensure wallet is not already unlocked: because of MV3 sleep mechanism, wallet might be unlocked on first call
+  const subscription = passwordStore.isLoggedIn.subscribe((isLoggedIn) => {
+    if (isLoggedIn === "TRUE") {
+      if (isLocked) sessionStore.reset()
+      subscription.unsubscribe()
```
alecdwm: Does this mean that if we unlock, lock, unlock, lock, unlock, that only the first unlock will trigger the sessionStore to be reset?

alecdwm (dismissed):

packages/extension-core/src/domains/app/store.session.ts
0xKheops: exactly.
will simplify this though: thanks to our keep alive system we can call reset straight up from background.ts, no need to tie this to the unlock

alecdwm (approved):

PR #1836, opened 2025-03-25
fix: upgrade `@polkadot/types` to fix signing error when extrinsic `assetId` is a `MultiLocation`

alecdwm:
In addition to a fix for the signing error, this PR also includes a change to throw when we fail to create a signed extrinsic, which will help to track down any similar problems we have in the future.


0xKheops (approved):
LGTM!

PR #1851, opened 2025-04-03
feat: discover more assets

0xKheops:
On each Asset Discovery scan, our Asset Discovery api will be called to query for owned tokens on 50+ chains. 
If tokens are found, associated chains will be scanned even if they are not default or forceScan chains


0xKheops:
PR updated to consider only regular ERC20s and native tokens. 
LP tokens wont be added anymore.

packages/extension-core/src/domains/assetDiscovery/registerMissingTokens.ts
```diff
@@ -0,0 +1,161 @@
...
+        await Promise.all(
+          assets.map(async (asset) => {
+            switch (asset.type) {
+              case "native": {
+                foundTokenIds.add(`${networkId}-evm-native`)
+                break
+              }
+              case "erc20": {
+                {
+                  const id = `${networkId}-evm-erc20-${asset.contractAddress.toLowerCase()}`
```
alecdwm: Could be cool to use `evmErc20TokenId` from `@talismn/balances` here. Also for the native token id above!

packages/extension-core/src/domains/assetDiscovery/scanner.ts
```diff
@@ -498,23 +534,22 @@ class AssetDiscoveryScanner {
...
+      // activate networks that have not been explicitely enabled or disabled
+      const networkIdsToActivate = evmNetworks.filter((n) => activeEvmNetworks[n.id] === undefined)
+      if (networkIdsToActivate.length) {
+        log.debug("[AssetDiscovery] Automatically enabling networks", networkIdsToActivate)
+        await activeEvmNetworksStore.set(
+          Object.fromEntries(networkIdsToActivate.map((t) => [t.id, true])),
+        )
       }
+
+      await sleep(100) // pause to ensure local storage observables fires before we exit
```
alecdwm: So something happens after we exit, and it needs to wait for the storage observables to have already fired?
What happens? Could be good to mention it here!

alecdwm (dismissed):
LGTM!

packages/extension-core/src/domains/assetDiscovery/scanner.ts
0xKheops: adjusted, thanks :)

packages/extension-core/src/domains/assetDiscovery/registerMissingTokens.ts
0xKheops: thanks, didnt realize those were exported!

0xKheops:
Edit: PR updated to only consider known tokens : the ones defined in chaindata, or added manually by the user
Rollbacked most changes, including the Custom token pills removal since we dont add new tokens to the db any anymore.

alecdwm (approved):
👌

PR #1862, opened 2025-04-09
fix: ledger issues

0xKheops:
- before signing, checks that the connected ledger has the expected address
- prevents sending funds to incompatible accounts (ex: sending AVAIL to a polkadot ledger account)
- prevents previous ledger error message to disappear if the payload changes (ie: fees updated)


alecdwm (approved):
:shipit:

PR #1869, opened 2025-04-11
fix: displays ledger error when checking address with a ledger legacy app

0xKheops:
When signing with a legacy app, at the check address step, if an error is raised it will be displayed properly instead of considering the address is missing.
For example if the ledger is locked, it will now display "Please unlock your ledger" instead of "Unable to normalize address: undefined"


alecdwm (approved):
:shipit:

PR #1871, opened 2025-04-11
Fix/bittensor decode dynamic info fallback

UrbanWill:
# Description

Calculate alpha token fiat value on the wallet frontend when decode dynamic data fails in the balances lib.

### 🗒️ **Notes:**
Subnet positions that failed to be decoded will not be included into TAO balance or total wallet balance.

### 🖼️ **Screenshots:**
In the pics below it know that subnet 64 is failing to decode, but the fiat token value is calculated on the FE:

Screenshot 2025-04-11 at 11 58 37
Screenshot 2025-04-11 at 12 01 49


0xKheops (approved):
LGTM

PR #1860, opened 2025-04-09
Fix/taostats api dtao update

UrbanWill:
# Description

Updated taostats api endpoint on `useGetBittensorInfiniteValidators` hook. This new endpoint does not offer APR (for now) so any code displaying/sorting APRs has temporarily commented out and APR is being displayed as "--" or "N/A"

### 🖼️ **Screenshots:**

Screenshot 2025-04-09 at 11 26 43

PR #1875, opened 2025-04-11
feat: rootnet validator apr

UrbanWill:
# Description

Adds new yield apr endpoint and pretty much reverts https://github.com/TalismanSociety/talisman/pull/1860

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/12c4f877-eca1-48df-a152-17e0ccdb663a

### DIFF
--- a/apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
+++ b/apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
@@ -26,6 +26,7 @@ import { TokenLogo } from "../../Asset/TokenLogo"
 import Tokens from "../../Asset/Tokens"
 import { TokensAndFiat } from "../../Asset/TokensAndFiat"
 import { STAKING_APR_UNAVAILABLE } from "../helpers"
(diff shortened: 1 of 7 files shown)


apps/extension/src/ui/domains/Staking/Bond/BondForm.tsx
```diff
@@ -320,16 +321,25 @@ const StakeApr = () => {
 
   const hookMap = {
```
0xKheops, line 322: this should be split into multiple components instead, hooks should not be called conditionally.
ok for now as it was already there but we should avoid this pattern in the future

0xKheops (approved):
LGTM if urgent.
But if not, would be great to fix the conditional hook calls before merging

0xKheops:
btw unrelated with this PR, but i see this when opening the stake modal
image

UrbanWill:
> btw unrelated with this PR, but i see this when opening the stake modal image

Thanks, I'll be getting rid of this hook when implementing staking on the wallet. (it is no longer required)

PR #1893, opened 2025-04-16
chore: playwright installation

prury:
PR summary: 
- Playwright installation
- CI configuration(building the wallet, installing dependencies, running playwright) (Headed mode)
- Added data-test ids trough some tags (please don't delete them, just modify accordingly if changing the code)
- Created a foundation for tests that will be written later, this foundation consists on: 
  1. Loading the wallet extension
  2. Making the browser context persistent so the tabs can communicate with the extension
  3. Automating the on boarding process so it doesn't need to be written again each test
  4. Automating basic account creation so its easy to create a new one
- scripts to run it locally can be found on the package.json

observations: headed mode is slower, but works best when used with wallet extensions. Headless mode has a hard time dealing with pop-ups and can be detected for being similar to bot behavior.


0xKheops (dismissed):
Amazing, LGTM!

alecdwm (dismissed):
LGTM 🔥🔥🔥🔥🔥 :shipit:

alecdwm (approved):

PR #1903, opened 2025-04-22
fix: keyring migration unexpected origins

0xKheops:
Fixes keyring migration for users that failed v1.19 migration


alecdwm (commented):
Should we add `HARDWARE` to `case LegacyAccountOrigin.Ledger`?

alecdwm (approved):
🚀

PR #1905, opened 2025-04-23
feat: ramps

0xKheops:


packages/extension-shared/src/constants.ts
```diff
@@ -32,6 +32,11 @@ export const BLOWFISH_API_KEY = process.env.BLOWFISH_API_KEY
 export const NFTS_API_KEY = process.env.NFTS_API_KEY
 export const NFTS_API_BASE_PATH = process.env.NFTS_API_BASE_PATH || "https://nfts-api.talisman.xyz"
 export const TAOSTATS_BASE_PATH = process.env.TAOSTATS_BASE_PATH || "https://tsp.talisman.xyz"
+export const RAMPS_COINBASE_API_BASE_PATH = "https://coinbase-api.talisman.xyz"
+// export const COINBASE_PROJECT_ID = "63080e24-dc8e-45d0-9618-467b8c222f9e"
```
alecdwm: `rm`?

apps/extension/src/ui/domains/Ramps/shared/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+const getEvmNativeTokenId = (evmNetworkId: string) => [evmNetworkId, "evm-native"].join("-")
+
+const getErc20TokenId = (evmNetworkId: string, contractAddress: string) =>
+  [evmNetworkId, "evm-erc20", contractAddress.toLowerCase()].join("-")
+
+const getDotSubstrateAssetTokenIdPrefix = (chainId: string, assetId: string) =>
+  [chainId, "substrate-assets", assetId].join("-")
+
+export const getTokenFromCoinbaseAsset = (asset: CoinbaseTokenNetwork, tokens: TokenList) => {
+  if (asset.chain_id && asset.chain_id !== "0") {
```
alecdwm: is `asset.chain_id` equal to `"0"` for polkadot and kusama? I would have assumed ethereum, in which case we're calling `getDotNativeTokenId` for ethereum assets?

apps/extension/src/ui/domains/Ramps/coinbase/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+  const remoteConfig = await remoteConfig$.getValue()
+
+  const query = new URLSearchParams({
+    appId: remoteConfig.ramps.coinbaseProjectId,
+    defaultAsset: assetId,
+    // defaultPaymentMethod: "CARD",
+    presetFiatAmount: amountIn,
+    fiatCurrency: currencyCode,
+    // defaultNetwork: network,
+    presetCryptoAmount: amountOut.toString(),
```
UrbanWill: This should be removed, since this will take precedence over `presetFiatAmount` and the users cannot choose cryptoAmount.

apps/extension/src/ui/domains/Ramps/shared/RampsTokenPicker.tsx
```diff
@@ -0,0 +1,234 @@
...
+      <div className="flex h-full min-h-full w-full flex-col overflow-hidden">
+        <div className="flex min-h-fit w-full items-center gap-8 px-12 pb-8">
+          <SearchInput ref={refSearchInput} onChange={setSearch} placeholder={t("Search")} />
+        </div>
+        <ScrollContainer
+          ref={refContainer}
+          className="bg-black-secondary border-grey-700 scrollable h-full w-full grow overflow-x-hidden border-t"
+        >
+          {!filteredTokens ? (
+            range(0, 10).map((i) => <TokenButtonRowSkeleton key={i} />)
```
alecdwm, line 107: This array could be memoised, but probably doesn't matter since it's just for the loading state!

UrbanWill (dismissed):
LGTM other than the minor comment on what to add to the buy url param.

Note: I could no test the offramp, it is broken on Coinbase side

apps/extension/src/ui/domains/Ramps/coinbase/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+    quoteId,
+    destinationWallets: JSON.stringify([{ address, blockchains: [network] }]),
+  })
+
+  return urlJoin(RAMPS_COINBASE_PAY_URL, `/buy/select-asset?${query}`)
+}
+
+// TODO check if it works
```
alecdwm: `rm`? 😀

apps/extension/src/ui/domains/Ramps/coinbase/useCoinbaseSellQuote.ts
```diff
@@ -0,0 +1,198 @@
...
+  currencyCode: string,
+  amountIn: number,
+  coinbaseToken: CoinbaseTokenSpecs,
+  decimals: number,
+  minMaxAmount: string | null,
+): Promise<FetchCoinbaseSellQuoteResult> => {
+  const body: CoinbaseSellQuoteRequest = {
+    cashoutCurrency: currencyCode,
+    paymentMethod: "FIAT_WALLET",
+    sellAmount: formatTokens(amountIn, decimals), // dont send "1e-9" ^^
```
alecdwm: `BigNumber(value).toFixed` also solves this!

> Unlike Number.prototype.toFixed, which returns exponential notation if a number is greater or equal to 1021, this method will always return normal notation.

From https://mikemcl.github.io/bignumber.js/#toFix

But your `formatTokens` also seems to do the trick

apps/extension/src/ui/domains/Ramps/ramp/useRampTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+import { RampsMode } from "../shared/types"
+import { getRampApiUrl } from "./getRampApiUrl"
+import { RampHostAssetsConfig } from "./types"
+import { useRampCurrencies } from "./useRampCurrencies"
+
+// note: currencyCode must be upper case
+const fetchRampAssets = async (
+  currencyCode: string,
+  mode: RampsMode,
+): Promise<RampHostAssetsConfig> => {
```
alecdwm: Maybe this should be a `/** */` comment, so it shows up in vscode whenever someone is using this function?

```suggestion
const fetchRampAssets = async (
  /** NOTE: currencyCode must be upper case */
  currencyCode: string,
  mode: RampsMode,
): Promise<RampHostAssetsConfig> => {
```

apps/extension/src/ui/domains/Ramps/shared/RampsCurrencyPicker.tsx
```diff
@@ -0,0 +1,204 @@
...
+      () => {
+        setIsIconsReady(true)
+      },
+    )
+  }, [])
+
+  // once drawer is open, focus on the search input
+  const refSearchInput = useRef<HTMLInputElement>(null)
+  const transitionStatus = useOpenCloseStatus()
+  useEffect(() => {
```
alecdwm, line 72: ~~I think `useLayoutEffect` might be appropriate here?~~

Nevermind! Re-reading https://react.dev/reference/react/useLayoutEffect, I'm mistaken.

alecdwm (dismissed):
Left some comments, but LGTM! We should merge this before swaps, since there'll be a few small conflicts.

apps/extension/src/ui/domains/Ramps/coinbase/helpers.ts
0xKheops: good catch, thanks

apps/extension/src/ui/domains/Ramps/shared/RampsTokenPicker.tsx
0xKheops: yeah we re in the case where we got no `tokens` prop, so it should render only once (if at all), should be acceptable

apps/extension/src/ui/domains/Ramps/shared/helpers.ts
0xKheops: yes it's "0" unless it's an EVM chain.
Though your right, maybe it could change in the future. will make it more bulletproof

alecdwm (approved):
:shipit:

Issue #1898, opened 2025-04-18
Unable to Send Transactions to Contract on Westend Network


smiasojed:
**Describe the bug**
Unable to Send Transactions to Contract on Westend Network Using Talisman Wallet

Transactions to a smart contract on the Westend network fail when using the Talisman Wallet. However, the same transactions work correctly when using MetaMask.
It appears that Talisman modifies the gas limit in a way that causes the transaction to fail.
We use a custom gas encoding mechanism to embed Polkadot-specific values in the transaction. This may be incompatible with how Talisman handles gas estimation or overrides it. More details about this encoding approach can be found here: https://github.com/paritytech/polkadot-sdk/pull/6689

**To Reproduce**
Steps to reproduce the behaviour:

1. Open Remix IDE (https://remix.polkadot.io) 
2. Compile `Storage` contract
3. Navigate to the Deploy & Run Transaction tab
4. In the Environment dropdown, select `Injected Provider – Talisman`.
5. If Talisman is not listed, click `Customize this list` and enable it.
6. Click `Deploy` button and approve the transaction
7. Find deployed contract and click `store` button with some value.

**Expected behaviour**
Transaction should pass.

**Environment (please complete the following information):**

- OS: [macOS]
- Browser [chrome]
- Chrome Version [135.0.7049.95]
- Talisman Version [2.7.5]

0xKheops:
Thank you for the heads up!

PR #1907, opened 2025-04-26
feat: polkadot evm support

0xKheops:
Testing steps:
- Open Remix IDE (https://remix.polkadot.io/)
- Compile Storage contract
- Navigate to the Deploy & Run Transaction tab
- In the Environment dropdown, select Injected Provider – Talisman.
- If Talisman is not listed, click Customize this list and enable it.
- Click Deploy button and approve the transaction
- Find deployed contract and click store button with some value.

=> With this PR, the transaction should be successful.

### DIFF
--- a/apps/extension/src/ui/domains/Ethereum/GasSettings/FeeOptionsForm.tsx
+++ b/apps/extension/src/ui/domains/Ethereum/GasSettings/FeeOptionsForm.tsx
@@ -9,14 +9,13 @@ import {
   EthTransactionDetails,
   GasSettingsByPriority,
   getTotalFeesFromGasSettings,
-  isAcalaEvmPlus,
 } from "extension-core"
 import { FC, useCallback, useMemo } from "react"
 import { Trans, useTranslation } from "react-i18next"
 import { Tooltip, TooltipContent, TooltipTrigger } from "talisman-ui"
 
 import { TokensAndFiat } from "@ui/domains/Asset/TokensAndFiat"
-import { useToken } from "@ui/state"
+import { useEvmNetwork, useToken } from "@ui/state"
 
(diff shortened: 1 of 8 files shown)


apps/extension/src/ui/domains/Ethereum/GasSettings/FeeOptionsForm.tsx
```diff
@@ -232,7 +232,7 @@ export const FeeOptionsSelectForm: FC<FeeOptionsSelectProps> = ({
             selected={priority === "recommended"}
           />
         )}
-        {!isAcalaEvmPlus(txDetails.evmNetworkId) && (
```
pgherveou, line 235: dumb question where do you define whether or not a chain has `preservveGasEstimate`

Metamask has a map here https://github.com/MetaMask/metamask-extension/blob/a420c18f0cc27289513a7c00d2af9b0a9d32cbd2/shared/constants/network.ts#L1112

but I can't find anything similar here

smiasojed:
This fixes: https://github.com/TalismanSociety/talisman/issues/1898, so it may be closed

apps/extension/src/ui/domains/Ethereum/GasSettings/FeeOptionsForm.tsx
0xKheops: hey @pgherveou 

Definitely not a dumb question :)

Its define in our chaindata repo, in which all chains are defined and can be updated without re-releasing the wallet.
The wallet syncs with it on startup and on regular basis.

https://github.com/TalismanSociety/chaindata

This will allow us to support kusama and polkadot asset hubs day one, as long as we change the flag there.

alecdwm (approved):
:shipit:

PR #1908, opened 2025-04-26
feat: rescue operations page

0xKheops:
Adds a new page (rescue.html) to Talisman.
It allows to backup and restore the local storage as a JSON file, which can be useful for both development and user-support purposes.

To navigate to it : 
- On login page, click 10 times the version number
- On onboarding's welcome page, click 10 times the Talisman logo

image

image

image


alecdwm (approved):
LGTM!

I think it might be worth obfuscating the backup file (e.g. base64 encoding the whole thing). It might help prevent the users who don't know what's going on in there from making problematic changes to it that later cause weird bugs and make us think we broke something when we didn't.

0xKheops:
as discussed on discord, being able to edit the storage can be useful in some cases.

PR #1910, opened 2025-04-28
fix: use lodash groupBy

UrbanWill:
# Description

Swap browser built in `groupBy` method to lodash's one.

Note to self: Check new supported methods with caniuse.com and Talisman's current supported min versions


0xKheops (approved):
LGTM!

PR #1914, opened 2025-04-29
fix: sync accounts catalog with keyring

0xKheops:
Fixes "unknown account" entries appearing in the popup.

### DIFF
--- a/packages/extension-core/src/domains/accounts/store.catalog.ts
+++ b/packages/extension-core/src/domains/accounts/store.catalog.ts
@@ -87,26 +87,6 @@ export class AccountsCatalogStore extends StorageProvider<AccountsCatalogData> {
     })
 
   /**
-   * This method should be called with any deleted addresses each time an account is removed from the keyring.
-   *
-   * This will ensure that the catalog and the keyring stay in sync.
-   */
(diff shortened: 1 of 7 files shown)


packages/extension-core/src/domains/accounts/store.catalog.ts
```diff
@@ -68,8 +72,19 @@ export class AccountsCatalogStore extends StorageProvider<AccountsCatalogData> {
...
+          !!recGetAllAddresses(tree)
+            .filter((ta) => !keyringAddresses.some((ka) => isAddressEqual(ta, ka)))
+            .filter((ta) => removeAccount(tree, ta)).length,
+      ).length
+
+      return hasAdded || hasRemoved
+    })
 
   /**
    * This method should be called with any deleted addresses each time an account is removed from the keyring.
```
alecdwm: Let's remove `removeAccounts` since we just use `syncAccounts` now?

0xKheops: oops yes my bad, thanks

alecdwm (approved):
:shipit:

PR #1919, opened 2025-04-30
fix: download images from statically if githack fails

0xKheops:
This one is just for you Ramo ♥️


alecdwm (approved):
Very nice! :shipit:

PR #1922, opened 2025-05-01
feat: enable default networks and tokens

0xKheops:
- When clicking enable all on networks page, user will now be prompter whether he wants to enable all recommended networks or all networks
- Also introduces "reset" buttons to reset active states to Talisman defaults on both networks and tokens pages. This is useful because Asset Discovery ignores tokens & networks that have explicitely been disabled.

Case to consider: 
- users adds vitalik.eth as watched account
- asset discovery enables a lot of tokens & networks
- if user removes the watched account, a lot of balances will keep being polled, until now there was no way to recover from this.
- reset buttons help recovering from this problem

Note: let's merge #1851 first, as it fixes the "scans triggering scans" bug, making it easier to test this


alecdwm (approved):
Awesome 🔥

PR #1924, opened 2025-05-02
chore: yeet spirit keys and gotp checks

0xKheops:


alecdwm (changes requested):
Looks like the tests are failing because we need to yeet spirit keys from the tests too

alecdwm (approved):

PR #1926, opened 2025-05-05
fix: swaps improvements

### LINKED ISSUE #1927: Swap improvements
- [x] Fix modal scrollbar height
- [x] Double check that we can swap TO tokens which the user doesn't have any balances for
- [x] Fix: flash of `Unsupported account type` in the confirm button
- [x] Fix fees to include commission
       https://github.com/TalismanSociety/talisman-web/issues/1335
- [ ] Figure out why BTC is not showing up as a recipient token
- [x] Speed up initial from token list (while fetching all stealthex routes)
- [ ] Show warning when recipient isn't a contact or watched address
- [ ] Remove reaping warning from the input field, show it as a drawer when clicking Swap like we do in send funds
- [ ] Don't show tokens in the recipient token list which don't have a route from the selected sender token
- [ ] https://github.com/TalismanSociety/talisman/issues/1899
- [ ] Convert token/account picker modals to drawers **make it slide 😎**
- [ ] ToAccount picker: show `X` as right-icon for any ToAccount which is not equal to FromAccount.
  - Clicking this icon will reset `ToAccount` back to the value of `FromAccount`.

alecdwm:
https://github.com/TalismanSociety/talisman/issues/1927


0xKheops (dismissed):
LGTM!

0xKheops (approved):

PR #1928, opened 2025-05-06
fix: incorrect typecasts to `EvmNetwork` in `src/ui`

alecdwm:
Our ui exclusively deals with `SimpleEvmNetwork`, not `EvmNetwork`:
```ts
export type SimpleEvmNetwork = Omit<
  EvmNetwork | CustomEvmNetwork,
  "balancesConfig" | "balancesMetadata"
>
```

This PR updates the types used in the UI, so that if we try to access the `EvmNetwork.balancesConfig` or `EvmNetwork.balancesMetadata` keys from the ui, typescript will let us know they're not available there.

I've also snuck in 3 uses of `combineLatest(Record)` instead of `combineLatest(Array).pipe(map(Array => Record))`, which makes the code a bit more concise.

### DIFF
--- a/.changeset/light-panthers-know.md
+++ b/.changeset/light-panthers-know.md
@@ -1,0 +1,6 @@
+---
+"@talismn/chaindata-provider": patch
+"@talismn/balances": patch
+---
+
+chore: add SimpleEvmNetworkList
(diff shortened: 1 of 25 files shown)


0xKheops (approved):
LGTM!

PR #1930, opened 2025-05-06
fix: allow multiple substrate-tokens to have the same symbol

alecdwm:
Fixes a bug where on chains like Hydration, we only fetch *one* of the three available USDC tokens (because they all have the same token id: `hydration-substrate-tokens-usdc`).

Tasks before merge:

- [x] Updated the [chaindata `overrideNativeTokenId` and `feeToken`](https://github.com/TalismanSociety/chaindata/blob/df6c9b9bf6eeefefcf76eea8ae026edf5674c311/data/chaindata.json#L2466-L2467) ids to consider the new id format.
- [x] Tested that the new chaindata `pub` files don't cause the existing prod wallet to crash (especially on the tx history screen).
- [x] Included a migration for tx history to update the token ids for existing transactions.
- [x] Included an update to [UNTRANSFERABLE_TOKENS](https://github.com/TalismanSociety/talisman/blob/b447fbf2930b9d76f95128df2e13543cced2dda1/apps/extension/src/ui/util/isTransferableToken.ts#L8C14-L8C35) to consider the new id format.
- [x] Included an update to the [api mocks](https://github.com/TalismanSociety/talisman/blob/b447fbf2930b9d76f95128df2e13543cced2dda1/apps/extension/tests/mocks/api.ts#L941) to consider the new id format.
- [x] Updated signet [supported tokens](https://github.com/TalismanSociety/signet-web/blob/8a1d2cdd6320477a059351e94dfbb0ed89194565/apps/multisig/src/domains/chains/all-tokens.ts) to consider the new id format.
  - https://github.com/TalismanSociety/signet-web/pull/105

### DIFF
--- a/.changeset/fresh-pets-whisper.md
+++ b/.changeset/fresh-pets-whisper.md
@@ -1,0 +1,5 @@
+---
+"@talismn/balances": patch
+---
+
+fix: allow multiple substrate-tokens to have the same symbol
--- a/apps/extension/src/ui/util/isTransferableToken.ts
+++ b/apps/extension/src/ui/util/isTransferableToken.ts
@@ -6,8 +6,8 @@ import { Token } from "@talismn/chaindata-provider"
 // ACA, BNC and KAR use native (orml won't work)
 // INTR, KINT and MGX use orml (native won't work)
 export const UNTRANSFERABLE_TOKENS = [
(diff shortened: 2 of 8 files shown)


0xKheops (approved):
LGTM!

PR #1940, opened 2025-05-08
feat: improved formatPrice and cleanup signature

0xKheops:
- always display at least 3 significant digits ($1 will display $1.00)
- added tests for formatPrice method
- removed the last argument of the method as it's not used anywhere (added it while working on ramps but didnt use it)


alecdwm (approved):
❤️

PR #1950, opened 2025-05-09
feat: vs TAO currency

alecdwm:
Adds TAO as a supported vs currency (just like we have with BTC, ETH, DOT).

Requires some macgyvering since TAO isn't a supported vs currency on coingecko.


0xKheops (approved):
Great job ser!

PR #1960, opened 2025-05-15
fix: connected accounts fix bundle

0xKheops:
- fixes the connected accounts pill so that the account name can use up to 50% (it was truncating after 1 character on websites with a long url)
- dev mode can now inject both watched-only and contacts properly (contacts were listed in the picker but wouldnt inject)
- for logic that injects account there is now a clear separation between dev mode (allow readonly & contacts) and portal infos => dev mode wont leak if an account is read-only or not.


alecdwm (approved):
LGTM :shipit:

PR #1963, opened 2025-05-17
fix: do not log webapp rpc calls

0xKheops:
webapp is using wallet to relay all its rpc calls.
all those calls were being logged, making it hard to use the log.
those wont be logged anymore.


alecdwm (commented):
:shipit:

alecdwm (approved):

PR #1976, opened 2025-05-22
feat: ledger polkadot secp256k1

0xKheops:
Prerequisite: Update to latest polkadot app on your Ledger at https://hub.zondax.ch/polkadot (this is not available yet on Ledger Live)

Changes:
- Ledger Polkadot app can now sign on Mythos and other secp256k1 account chains, as long as they have the CheckMetadataHash signed extension.
- Ledger Substrate Legacy apps now have a custom derivation path UI (some users were complaining that this was only available for the generic app)
- Ensure `ledger-ethereum` accounts can't be injected on substrate dapps, and that `ledger-polkadot` accounts can't be injected on evm dapps

Things to know: 
- Onboarding a Ledger Polkadot app account on chains that use EVM addresses will create a different account from than if it was on a substrate format address chain. => I changed the default naming of accounts from `Ledger Polkadot X` to `Ledger {chainName} X`
- There is a migration on first startup, to add the new `curve` property on ledger polkadot accounts. possible values are `ed25519` or `ethereum`
- Those new accounts, despite having an ethereum address, can't sign on EVM networks. It's substrate only.
- It allows allow signing on the substrate side of moonbeam, but not on the EVM side. Additionally, adding the same account using Ledger ethereum app wouldn't be possible as it would result in the same address, which serves as ID in our keyring (cant have duplicates).

Testing grounds:
- Kheopswap on [Operation branch](https://feat-operation.kheopswap-xyz.pages.dev/#/polkadot/operation) => can transfer MYTH on mythos chain, and on AH to test for no regression on regular ledger generic accounts.
- Mythos PJS app [sign message](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fmythos.dotters.network#/signing) and [verify signature](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fmythos.dotters.network#/signing/verify)
- Polkadot AH PJS app for regression tests on regular ledger generic accounts [sign message](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fasset-hub-polkadot-rpc.dwellir.com#/signing) and [verify signature](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fasset-hub-polkadot-rpc.dwellir.com#/signing/verify)

@alecdwm @UrbanWill  : This PR includes changes to the keyring account typeguards:
- `isAccountEthereum` => renamed to `isAccountAddressEthereum` checks if an account has an EVM address
- `isAccountPlatformEthereum` checks if an account is compatible with EVM networks (which is not the case for Ledger Polkadot secp256k1 accounts). 

same on polkadot side:
- `isAccountPolkadot` => renamed to `isAccountAddressSs58` checks if an account has an SS58 address
- `isAccountPlatformPolkadot` checks if an account is compatible with Substrate chains (allows both ss58 and ethereum format accounts, but excludes Ledger Ethereum accounts as they cant sign there).


packages/extension-core/src/domains/accounts/helpers.ts
```diff
@@ -137,8 +138,21 @@ export const isCurveCompatibleWithChain = (
...
+    : chain.account !== "secp256k1"
+}
+
+export const isAccountCompatibleWithNetwork = (
+  network: SimpleEvmNetwork | Chain,
+  account: Account,
+) => {
+  return /^\d+$/.test(network.id)
+    ? isAccountPlatformEthereum(account)
+    : isAccountCompatibleWithChain(network as Chain, account)
```
alecdwm: I would strongly prefer if we didn't start testing whether chain/network ids are strings or numbers 😢
Also I don't think this function is currently used anywhere!

```suggestion
export const isAccountCompatibleWithNetwork = (network: SimpleEvmNetwork, account: Account) =>
  isAccountPlatformEthereum(account)
```

packages/extension-core/src/domains/keyring/migrations/migrateLedgerPolkadotCurve.ts
```diff
@@ -0,0 +1,9 @@
+import { Migration, MigrationFunction } from "../../../libs/migrations/types"
+import { keyringStore } from "../store"
+
+// keyring will automatically add the missing curve property to ledger-polkadot accounts
+// just need to call forceUpdate() for the keyring's storage to be updated
```
alecdwm, line 5: Awesome

packages/extension-core/src/domains/accounts/helpers.ts
0xKheops: good catch, forgot to delete it :)

0xKheops: would be nice to have such a helper to be able to check if an account (which can be evm or substrate) can sign on a network (which can be evm or substrate), but it can wait until networks consolidation update.

alecdwm (approved):
:shipit:

PR #1979, opened 2025-05-24
fix: dashboard layout truncated main area

0xKheops:
Solves a problem where in the dashboard some content is truncated and not reachable via scrolling, on very small resolutions.
In this example, the bottom of the form (ledger connection status and continue button) is not displayed at all and scrolling wont help:
image


alecdwm (approved):
LGTM!

PR #1984, opened 2025-05-26
feat: stealthex swaps

alecdwm:
- [x] add stealthex swap module
- [x] update existing swaps code to enable switching swap modules


0xKheops (approved):
LGTM!

PR #1985, opened 2025-05-27
chore: package maintenance

0xKheops:
- moves all non-wallet build related dependencies to devDependencies, allowing us to leverage `pnpm audit --prod` without bloat
- removes nanoid dependency
- bumps dompurify and @ledgerhq/hw-app-eth dependencies, which were flagged as having vulnerable dependencies by pnpm audit
- applied `pnpm audit --prod --fix`


alecdwm (dismissed):
<div align="center">

radical

</div>

alecdwm (approved):

PR #1989, opened 2025-05-28
feat: polkadot vault v7 tx signing with metadata proof

0xKheops:


apps/extension/src/ui/domains/Sign/Qr/SignPayloadQrCode.tsx
```diff
@@ -0,0 +1,90 @@
+import { wrapBytes } from "@polkadot/extension-dapp/wrapBytes"
+//import { createSignPayload } from "@polkadot/react-qr/util"
```
alecdwm: `rm`?

alecdwm (dismissed):
LGTM :shipit:

apps/extension/src/ui/domains/Sign/Qr/SignPayloadQrCode.tsx
0xKheops: oops, thanks :)

alecdwm (approved):

PR #2021, opened 2025-06-06
feat: dry runs in send funds

0xKheops:
dry runs send funds transactions to prevent submitting transactions that would fail.

for example, if trying to send MYTH to an account that is not active (has no sufficient assets), as MYTH isnt a sufficient asset itself, the transaction would fail.

For QA might want to simulate transactions on all substrate chains/assets to see if adding dry runs doesnt cause any issue


0xKheops (commented):
> dry runs send funds transactions to prevent submitting transactions that would fail.
> 
> for example, if trying to send MYTH to an account that is not active (has no sufficient assets), as MYTH isnt a sufficient asset itself, the transaction would fail.

alecdwm (dismissed):
Nice 👌

prury (approved):

PR #2023, opened 2025-06-06
chore: bump to papi and pjs latest

0xKheops:
both papi and pjs have upgraded to Metadata V16
Updating broke a bunch of bits, especially in balance libraries and mini metadata, that were only accounting for metadata v14 and v15. Had to make several adjustments.


0xKheops (commented):
chore: changeset

alecdwm (approved):
LGTM!

PR #2025, opened 2025-06-09
fix: include all fees in swaps fee display

alecdwm:
Previously we only included gas fees in the `Fees` section, now we include everything


0xKheops (approved):
LGTM!

PR #2029, opened 2025-06-10
fix: swaps fee calculation when fromAsset is ETH

alecdwm:
`BigNumber(10).pow(-fromAmount.decimals)` loses precision, whereas `10 ** -fromAmount.decimals` maintains it (and is what we already use in our `tokensToPlanck`/`planckToTokens` utils).

When swapping from ETH (decimals 18), this was enough loss of precision to round to `$0.00` for the fees display!


0xKheops (approved):
LGTM!

PR #2032, opened 2025-06-10
feat: consolidate fetch metadata method

0xKheops:
adds a `fetchBestMetadata` method in sapi that is used by both the extension and balance libraries


alecdwm (approved):
🔥

PR #2043, opened 2025-06-12
feat: chaindata v4

0xKheops:
Read chaindata-v4 changes first, this PR depends on it: https://github.com/TalismanSociety/chaindata/pull/203

Changes
- Temporary backend subscriptions (ex: `useBalancesByParams`) can now be unsubscribed  - this was copy/pasted from the pending NFTs PR
- `Chain` and `EvmNetwork` types have been removed in favor the new `Network` type
- `ChaindataProvider` interface has been refactored to account for the above
- ⚠️ The Portal/Signet-specfic injected object now returns the new `Network` objects instead of `Chain` and `EvmNetwork` objects : custom networks and custom tokens endpoints will now throw an error
- ⚠️ Token ids changed, they now only contain what identifies a token on chain, they also use a different separator (`-` => `:`) so they can be parsed/reversed
- networks are now a single entity and db table, with a platform discriminator property to distinguish between `DotNetwork` and `EthNetwork` types
- custom networks and tokens are not stored in chaindata’s db anymore, but in the local storage of the wallet
- ⚠️ Networks and Tokens are typechecked with Zod, and are used as-is by chaindata => from now on if types are not backward compatible, need to bump chaindata version (pub/v*). Basically, any change other than adding an optional property requires a chaindata bump
- `SubEquilibriumToken` module has been deleted
- Crowdloans module has been deleted
- Balance modules now always check chain’s spec version (and update if necessary) before reusing minimetadatas from db
- 🔥 Balance modules now process balance subscriptions network by network, allowing to fetch mini metadata on demand without having all networks subs wait for mini metadatas to be ready
- `ChaindataProvider` doesnt need to be “hydrated” anymore, and it will self update from github every 5 mins. Though list of custom tokens/networks can still provided via it’s constructor
- `isTestnet` is not a token property anymore, though a new `isTokenTestnet(token)` helper achieves the same result. This will however only work if the token is issued from a chaindataProvider (which matches it with networks to check if they are testnets or not)
- `isCustom` is not a token/network property anymore, though new `isTokenCustom(token)` and `isNetworkCustom(network)` helpers achieve the same result. This will however only work if the token/network is issued from a chaindataProvider (which merges github chaindata and custom networks/tokens together)
- `token.evmNetwork?: {id:string}` and `token.chain? : {id:string}` are replaced by `token.networkId: string`
- on DotNetwork, `relay?:{id:string}` and `paraId` are moved to the discriminable `topology` property
- tokens have a `name` property
- `@talismn/chaindata-provider` includes a bunch of new type helpers for networks and tokens `isTokenOfType`, `isNetworkOfPlatform`, `isTokenErc20`, `isNetworkDot`  etc. and the official methods (found and removed some copies in multiple places) to generate a tokenId: `evmErc20TokenId`, `subPsp22TokenId`, etc.
- Networks page displays all networks in one list
- Tokens page displays all tokens in one list
- All types of tokens now have a "token details" page and users can adjust symbol and name there
- In asset detail page, added an entry to link the token details from the token context menu
- for frontend developement, `useChains(id)` and `useEvmNetwork(id)` have been replaced by `useNetwork(id, type?)`
- On the tokens page, the "enable all" button has been removed
- Token rates are instantly reloaded if user currencies list change (ex: adding TAO)
- Send funds's token picker shouldn't hang while loading anymore
- Reworked UI of add & edit contact forms (replaced the network dropdown by a modal picker)
- Fix: allow removing the specific network on contacts
- ChainLogo is now NetworkLogo
- Refactored AssetLogo (takes an url) and TokenLogo (takes a token id)


packages/balances-react/src/atoms/chaindataProvider.ts
```diff
@@ -1,35 +1,19 @@
...
 
-export const miniMetadataHydratedAtom = atom(false)
-
 /** This atomEffect keeps chaindata hydrated (i.e. up to date with the GitHub repo) */
-const chaindataHydrateAtomEffect = atomEffect((get, set) => {
-  const chaindataProvider = get(chaindataProviderAtom)
-  const miniMetadataUpdater = get(miniMetadataUpdaterAtom)
-  const evmTokenFetcher = get(evmTokenFetcherAtom)
-
+const chaindataHydrateAtomEffect = atomEffect((get, _set) => {
```
alecdwm: Look to me like ChaindataProvider is self-hydrating now - in which case, we should delete this atom rather than keep it as a useless placeholder

packages/balances/src/modules/SubstrateTokensModule.ts
```diff
@@ -160,51 +151,85 @@ export const SubTokensModule: NewBalanceModule<
...
+              chaindataProvider,
+              addressesByToken,
+              controller.signal,
+            )
+            if (controller.signal.aborted) return () => {}
+
+            const stateHelper = new RpcStateQueryHelper(chainConnector, queries)
+
+            return await stateHelper.subscribe((error, result) => {
+              //  console.log("SubstrateAssetsModule.callback", { error, result })
```
alecdwm:
```suggestion
```

alecdwm (dismissed):

packages/balances-react/src/atoms/chaindataProvider.ts
0xKheops: sounds good!

alecdwm (approved):
<div align="center">

LGTM

</div>

PR #2075, opened 2025-07-03
feat: hydration tokens

0xKheops:
Goes along with chaindata PR: https://github.com/TalismanSociety/chaindata/pull/205

- New dumbified balance module interface
- Each balance module is now a folder, meant to be copy/pastable to create new ones easily
- BalanceProvider class: the new entry point for subscribing to balances: it provides a getBalances$() method which returns an observable, which concurrently fetches balances for each network, after fetching their minimetadata.
- BalanceProvider's storage has to be handled by it's consumer. Wallet stores it as a blob in the main wallet db. Balances library dedicated db is deleted.
- New balance module for hydration tokens (which polls runtime calls)
- Fixes native modules locked balance values that were sometimes out of sync with the main subscription
- Native balances module creates less subscriptions
- Networks, tokens and accounts can now be enabled/disabled/added without causing unimpacted balance subscriptions to restart. each network subscription is independant.
- New CLI project in the monorepo allowing to test balance modules: `./apps/balances-bench`
- Frontend balances subscription can now be unsubscribed to without closing the tab, for example when the user navigates to settings (after 3 seconds)
- Balances & portfolio rxjs hooks do not suspense anymore (prerequisite for the above)
- Balances fetched via useBalancesByParams are kept in the same cache than wallet balances, meaning that after adding an account with a preview of balances, those balances are immediately shown in portfolio
- Transfer payloads are now crafted from the front end
- New "transfer" `txInfo` type, and deprecate legacy untyped `txInfo`
- New generic TxSubmitButton component, that can sign anything with any type of account
- Deleted the balance totals backend service, displayed totals are now all computed on frontend => totals are now always in sync with the rest of the UI
- Fixes Asset Discovery which wasn't working properly since Chaindata v4 PR
- Display token names in portfolio and token picker
- Replaced network dropdown by a new network combobox in tokens list page and add token page

Issues list https://github.com/TalismanSociety/talisman/issues/2082

PR Review soundtrack: https://open.spotify.com/track/4oaZRvZMytQvISa6A0ZKx0


packages/on-chain-id/src/util/addressesToNames.ts
```diff
@@ -89,13 +89,12 @@ export const lookupEnsAddresses = async (
         const domain = await client.getEnsName({ address })
         domain !== null && onChainIds.set(address, domain)
       } catch (cause) {
-        throw new Error(`Failed to resolve ENS domain for address '${address}': ${String(cause)}`, {
-          cause,
-        })
+        const errorMessage = (cause as { shortMessage?: string }).shortMessage ?? String(cause)
```
alecdwm:
```suggestion
        const errorMessage = (cause as { shortMessage?: string })?.shortMessage ?? String(cause)
```

alecdwm (dismissed):
`@talismn/balances` has never looked this good 🔥🔥🔥

0byt (approved):

PR #2091, opened 2025-07-15
feat: solana

0xKheops:
⚠️  The chaindata https://github.com/TalismanSociety/chaindata/pull/206 needs to be merged before this one

Misc. changes:
- Replaced deprecated bytesToString and stringToBytes calls with proper encoding/decoding methods
- Rewritten the add account from private key screen with `@tanstack/react-forms`
- Replaced our p.js based encodeAnyAddress/convertAddress helpers by ones from `@talismn/crypto` that support solana addresses too
- Solana support for: keypair and ledger accounts, native & SPL balances, send funds, dapp injection, asset discovery
- Migrated transactions db table to a new transactionsV2 table, with a new structure and a platform discriminator
- Improved account creation consistency by not using "SURI" anymore, but mnemonic + derivation path instead, for all account platforms
- Updated p.js libs to 13.5.3 to avoid warnings, related to the `@polkadot/crypto-utils` recent upgrade

Impacts on existing screens / QA scope:
- Onboarding, especially with mnemonic and/or private key - major changes there
- Copy address
- Send funds (especially the recipient picker)
- Tx History
- Contacts
- Authorized Sites
- Transactions from dapps 
- Ledger (all libs had to be upgraded including substrate and ethereum)
- Account select/change when connected to a dapp + might want to check behavior on dapps that connect to multiple providers (eth+sub, eth+solana, sub+solana)
- Anything else that involves addresses really (since validation helpers changed)
- Signing on Astar (which uses a tip) with send funds


packages/util/src/index.ts
```diff
@@ -1,12 +1,7 @@
 export * from "./addTrailingSlash"
 export * from "./BigMath"
-export * from "./blake2Concat"
 export * from "./classNames"
-export * from "./convertAddress"
-export * from "./decodeAnyAddress"
-export * from "./decodeSs58Format"
 export * from "./deferred"
-export * from "./encodeAnyAddress"
 export * from "./firstThenDebounce"
```
alecdwm, line 5: encodeAnyAddress/decodeAnyAddress is still used in Portal, so we'll need to remember to switch it to use the new one when we upgrade to the latest version of `@talismn/util`.

packages/solana/src/utils/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./serialization"
+export * from "./signing"
+export * from "./chains"
+export * from "./transaction"
```
alecdwm: I think instead of having `index.js` in each directory, we should just have one `index.js` in the package root.

The root one can then have:

```ts
export * from './utils/serialization'
export * from './utils/signing'
export * from './utils/chains'
export * from './utils/transaction'
```

The (only) advantage of this, is it prevents circular dependencies:

If `signing` imports functions from `chains` (as an example), it can only import them from `./chains`.

Whereas with the `index.js` in this directory, it can import them from either `./chains` or `.`.

Since `.` also imports `signing`, we end up with the beginnings of a circular dependency setup.

packages/solana/src/utils/chains.ts
```diff
@@ -0,0 +1,23 @@
+export const SOLANA_CHAINS = [
+  "solana:mainnet",
+  "solana:devnet",
+  "solana:testnet",
+  "solana:localnet",
+]
```
alecdwm: If we add `as const` here then `typeof SOLANA_CHAINS` becomes `"solana:mainnet" | "solana:devnet" | etc` instead of `string`. If that's what we want!

```suggestion
] as const
```

.changeset/config.json
```diff
@@ -2,10 +2,6 @@
   "$schema": "https://unpkg.com/@changesets/config@2.1.1/schema.json",
   "changelog": "@changesets/cli/changelog",
   "commit": false,
```
alecdwm, line 4: Can we keep the `@talismn/balances` & `@talismn/balances-react` versions in sync?

```suggestion
  "commit": false,
  "fixed": [
    ["@talismn/balances", "@talismn/balances-*"]
  ],
```

packages/chaindata-provider/src/state/net.ts
```diff
@@ -28,7 +28,7 @@ const fetchJsonFromGithubUrl = async <T>(
   const req = await fetch(url, { signal })
 
   // uncomment the line below to test initChaindata provisioning
-  // if (Date.now()) throw new Error("OMG HE GOT A KNIFE!")
+  // if (Date.now()) throw new Error("OMG SHE GOT A KNIFE!")
```
alecdwm, line 31: 😆

packages/chaindata-provider/src/state/initChaindata.json
```diff
@@ -4,14 +4,14 @@
       "id": "1",
       "isDefault": true,
       "name": "Ethereum Mainnet",
-      "logo": "https://raw.githubusercontent.com/TalismanSociety/chaindata/main/assets/chains/1.svg",
+      "logo": "https://raw.githubusercontent.com/TalismanSociety/chaindata/feat/analos/assets/chains/1.svg",
```
alecdwm: TODO: Change back to main before mergin

packages/chaindata-provider/src/state/defaultChaindata.ts
```diff
@@ -104,6 +106,10 @@ export const defaultChaindata$ = new Observable<Chaindata>((subscriber) => {
           "rw",
           ["networks", "tokens", "miniMetadatas"],
           async (ctx) => {
+            // we may end up here if db data is invalid, so we need to clear the tables first
+            await ctx.tokens.clear()
+            await ctx.miniMetadatas.clear()
+            await ctx.networks.clear()
```
alecdwm: Might be good to remove the changes to this file when merging in the latest `dev`, since it all changed with the dexie -> blob changes!

.changeset/config.json
0xKheops: sorry my bad, thought it was a leftover from the old per-module balance libraries

apps/extension/src/ui/state/networks.ts
```diff
@@ -33,6 +33,8 @@ export const [useNetworkDisplayTypesMapById, networkDisplayTypesMapById$] = bind
               const dotNetwork = network as DotNetwork
               return [network.id, getDotNetworkType(dotNetwork, networksById, t)]
             }
+            case "solana":
```
alecdwm, line 36: TODO: Port this into the Portal copy of this function
(just a reminder for myself)

apps/extension/src/ui/apps/popup/pages/Connect.tsx
```diff
@@ -180,7 +192,7 @@ export const ConnectPolkadot: ConnectComponent = ({
         </ConnectAccountsContainer>
         {!accounts.length && (
           <NoAccountWarning
-            type={"polkadot"}
+            type={"solana"}
```
alecdwm: Should this still be "polkadot"?

apps/extension/src/ui/hooks/sapi/useScaleApi.ts
```diff
@@ -26,7 +26,7 @@ export const useScaleApi = (
       if (!chain?.genesisHash || !token) return null
 
       const metadataDef = await api.subChainMetadata(chain.genesisHash, specVersion)
-      assert(metadataDef?.metadataRpc, `Metadata unavailable for chain ${chain.id}`)
+      assert(metadataDef?.metadataRpc, `Failed to fetch metadata`)
```
alecdwm, line 29: Do we definitely want to remove the chain.id from this error message? I'm thinking it might help debug which chain is causing problems if we ever see it

alecdwm (dismissed):
**L F GEEEE**

packages/chaindata-provider/src/state/defaultChaindata.ts
alecdwm: All sorted by [0975c52](https://github.com/TalismanSociety/talisman/pull/2091/commits/0975c526280621091c5873a5c1e0ca3811a31582)

apps/extension/src/ui/hooks/sapi/useScaleApi.ts
0xKheops: the intent was to make the message more concise, as this bubbles up to send funds UI when metadata cant be fetched

0xKheops: iirc when this happens it also logs an error that includes the chain id on the backend side

packages/solana/src/utils/index.ts
0xKheops: sounds good!

apps/extension/src/ui/hooks/sapi/useScaleApi.ts
alecdwm: Sounds good!

packages/chaindata-provider/src/state/initChaindata.json
0xKheops: done :)

alecdwm (approved):
<div align="center">

activate

</div>

PR #2108, opened 2025-07-17
fix: inject scripts

0xKheops:
removes all the dependencies from the extension-shared packages, which is meant to export only constants and types used by both the extension and the extension inject scripts


alecdwm (approved):
👌

PR #2120, opened 2025-07-31
feat: swappable chaindata storage

alecdwm:
- Added `TalismanChaindataV4` to the list of DBs to delete in our error boundary
- Removed new `chaindata` blob from list of blobs to be cleared by `Reload Talisman` button
- Removed some `@talismn/balances` code we don't use anymore
- Removed some barrel files from `@talismn/chaindata-provider`
- Added `storage$` argument to `ChaindataProvider` for injectable persistence handling
- Refactored `defaultChaindata$` to use `$storage` for persistence
- Added code to delete `TalismanChaindataV4` dexie db if present
- Added `store.chaindata.ts` for blob chaindata persistence in wallet
- Made `store.balances.ts` more consistent with `store.chaindata.ts` since they provide similar functionality
- Extracted tokenRates db into swappable storage

---

- [ ] Extract connectionMetaDb into swappable storage


packages/chaindata-provider/src/state/oldDb.ts
```diff
@@ -0,0 +1,34 @@
...
+    })
+  }
+}
+
+const oldChaindataDb = new OldChaindataDb()
+
+export const tryToDeleteOldChaindataDb = () => {
+  try {
+    // try and delete it if it's still there
+    oldChaindataDb.delete()
```
0xKheops: maybe can yeet the file and just do indexedDB.deleteDatabase("TalismanChaindataV4") ?

packages/chaindata-provider/src/provider/ChaindataProvider.ts
```diff
@@ -1,51 +1,68 @@
...
+
+/** By default, chaindata will be stored in memory and not persisted. */
+const DEFAULT_STORAGE: ChaindataStorage = {
+  networks: [],
+  tokens: [],
+  miniMetadatas: [],
+}
 
 export type ChaindataProviderOptions = {
+  storage$?: Subject<ChaindataStorage> | ChaindataStorage
```
0xKheops: worried about storage$ being passed as an observable, meaning the caller could change its value.
unlike custom$ which needs to be updated by the app/wallet afterwards based on user actions, i think `default` should be a simple object that is used only as init data, like for the balances provider.

in my mind it would only be used like this: `const defaultChaindata$ = githubChaindata$.pipe(startWith(storage))`
wdyt ?

prury:
I think this is a legit playwright detection: 
image

packages/chaindata-provider/src/state/defaultChaindata.ts
```diff
@@ -1,174 +1,108 @@
...
+            miniMetadatas: initChaindata.miniMetadatas as Chaindata["miniMetadatas"],
+          })
+
+          log.info("[defaultChaindata$] Initial chaindata file imported successfully")
+        } catch (cause) {
+          log.error("[defaultChaindata$] Failed to import initial chaindata file", { cause })
+          return
+        }
+      },
+      next: async (githubData) => {
```
0xKheops, line 52: this could be dumbified: now that it's stored as a blob, we don't need to sort & compare individual collections anymore.

packages/chaindata-provider/src/state/defaultChaindata.ts
```diff
@@ -1,174 +1,108 @@
...
+
+        try {
+          // if fetching from github fails, and if DB is empty, provision it with initial data
+          log.info("[defaultChaindata$] Importing initial chaindata file", initChaindata)
+
+          storage$.next({
+            networks: initChaindata.networks as Chaindata["networks"],
+            tokens: initChaindata.tokens as Chaindata["tokens"],
+            miniMetadatas: initChaindata.miniMetadatas as Chaindata["miniMetadatas"],
+          })
```
0xKheops: maybe just `storage$.next(initChaindata)`

0xKheops (approved):
LGTM! 🔥

PR #2139, opened 2025-08-15
feat: auto detect new solana tokens

0xKheops:
- balances provider now exposes a getDetectedTokens$(address) observable
- detected tokens will be automatically enabled by the solana asset discovery module

Note: this only works for Solana SPL tokens for now, but it is not balance module specific. in the future more balance modules may leverage it too (ex: Token2022)


alecdwm (approved):
👌

PR #2154, opened 2025-09-02
fix: crash on tx history due to swaps status

alecdwm:
After failing to fetch the status from the upstream API, we'll now default to an `Unknown` status instead of throwing an error.
Prevents the UI from crashing due to that unhandled error.

I also took the opportunity to tidy up the cache & retry & refresh logic using Observables, because I noticed a bug with the existing cache mechanism and it was easier to just comprehensibly rewrite it than to try and fix the existing bug.

Also long term we wanna use rxjs instead of jotai

### DIFF
--- a/apps/extension/src/ui/domains/Swap/swap-modules/simpleswap-swap-module.ts
+++ b/apps/extension/src/ui/domains/Swap/swap-modules/simpleswap-swap-module.ts
@@ -12,9 +12,19 @@ import { ScaleApi } from "@talismn/sapi"
 import BigNumber from "bignumber.js"
 import { remoteConfigStore } from "extension-core"
 import { UNKNOWN_TOKEN_URL } from "extension-shared"
-import { atom, ExtractAtomValue, Getter } from "jotai"
-import { withAtomEffect } from "jotai-effect"
+import { atom, ExtractAtomValue } from "jotai"
 import { atomWithObservable, loadable } from "jotai/utils"
+import {
+  catchError,
+  defer,
+  interval,
+  Observable,
+  of,
+  retry,
+  startWith,
+  switchMap,
+  takeWhile,
+} from "rxjs"
 import { encodeFunctionData, erc20Abi, publicActions, TransactionRequest } from "viem"
 import {
   arbitrum,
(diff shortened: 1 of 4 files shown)


apps/extension/src/ui/domains/Swap/swap-modules/simpleswap-swap-module.ts
```diff
@@ -884,37 +894,35 @@ export const simpleswapSwapModule: SwapModule = {
...
+const retryStatus$ = (id: string): Observable<Exchange["status"] | undefined> =>
+  defer(() => simpleSwapSdk.getExchange(id).then((exchange) => exchange.status)).pipe(
+    // retry up to 10 times, wait 5s between each retry
+    retry({ count: 10, delay: 5_000 }),
+
+    // log when all retries failed, and return undefined
+    catchError((error) => {
+      // eslint-disable-next-line no-console
+      console.error(`Failed to fetch exchange status for '${id}'`, error)
+      return of(undefined)
```
alecdwm, line 926: This is the core of the fix - we return undefined (which is rendered as `Unknown` in the UI) instead of throwing at this point!

apps/extension/src/ui/domains/Swap/swap-modules/stealthex-swap-module.ts
```diff
@@ -861,37 +871,35 @@ export const stealthexSwapModule: SwapModule = {
...
+const retryStatus$ = (id: string): Observable<StealthexExchange["status"] | undefined> =>
+  defer(() => stealthexSdk.getExchange(id).then((exchange) => exchange.status)).pipe(
+    // retry up to 10 times, wait 5s between each retry
+    retry({ count: 10, delay: 5_000 }),
+
+    // log when all retries failed, and return undefined
+    catchError((error) => {
+      // eslint-disable-next-line no-console
+      console.error(`Failed to fetch exchange status for '${id}'`, error)
+      return of(undefined)
```
alecdwm, line 903: Here too, for stealthex

0xKheops (approved):
awesome hidden gems :)

PR #2158, opened 2025-09-08
fix: drawer crash

UrbanWill:


UrbanWill:
Same as: https://github.com/TalismanSociety/talisman/pull/2193

PR #2171, opened 2025-09-18
chore: force eslint version also for peerDependencies

alecdwm:
Apparently `pnpm.overrides` doesn't apply to `peerDependencies`, unless they're also listed in `dependencies`/`devDependencies`.

This PR adds eslint v8 to `devDependencies` so we can override it in the packages that specify it as a `peerDependency`, until such a time that we switch to eslint v9.


0xKheops (approved):
My hero ❤️

PR #2172, opened 2025-09-20
chore: bump project to nodejs v24

alecdwm:
Versions bumped:
- `engines.node` in all `package.json` files to `>=20`
- `.nvmrc` to `v24.8.0`
- `use-node-version=24.8.0` in `.npmrc`
- `@types/node` bumped to `^24.5.1`
- `FROM node:24.8.0` in `Dockerfile` used for reproducible firefox builds

Broken things (due to the above bump) fixed by this PR:
- `@commitlint/cli` bumped to latest version to fix crash when committing with node v24
- `i18next` and friends bumped to latest version to fix deprecation warning when downloading translations
- `jest-environment-jsdom` extended to include `global.CryptoKey` from nodejs, which in v24 has been moved from `global.crypto.webcrypto.CryptoKey` to `global.CryptoKey`

Tested:
- All `package.json` scripts
- Wallet dev builds load & can sign txs in the browser
- Wallet prod builds

To note: it's good idea to run `nvm install` to ensure you're using the new node version.
If you have node v18 installed globally, you'll still drop into it when calling `node` on the CLI, which might lead to some confusion.


0xKheops (approved):
LGTM!

PR #2177, opened 2025-09-29
chore: use less memory when minifying js

alecdwm:
To test this, I tried running the firefox build in an environment which only has access to 4GB of memory.

On `dev`, this causes it to exit with code 137 (out of memory).

On this branch, it builds successfully.


0xKheops (approved):
LGTM

PR #2192, opened 2025-10-07
feat: swaps fixes

alecdwm:
- Fixed *all* swaps failing to load routes when *any single* swapModule fails to load its routes
- Reduced hardcoded eth gas buffer (for tx submission) and included swaps gas limit in buffer calculation (for lifi)
- Pull talismanTokens (for lifi) from config
- Included LIFI builtin fee in fee display
- Removed talisman fee from talismanTokens


0xKheops (approved):
LGTM!

PR #2193, opened 2025-10-08
Feat/seek staking discounts bittensor

UrbanWill:
# Description

Depends on: https://github.com/TalismanSociety/talisman-config/pull/14

Added Bittensor dTao staking discounts for users staking SEEK.

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/870e7afc-0c0a-4641-aeb4-1816442f8595

PR #2197, opened 2025-10-09
fix: staking buttons visibility

0xKheops:
previously the staking button and hooks were passed a group of balances (grouped by symbol) plus an arbitrary tokenId as parameter, which was used as reference. if that tokenId wasnt the one of a stakable token, then the button wasnt showing even if some other balances were stakable.

=> removed the tokenId parameter, the button will display if any of the balances is stakable.


apps/extension/src/ui/domains/Staking/Bond/hooks/useBondButton.ts
```diff
@@ -1,165 +1,212 @@
...
+        case "bittensor": {
+          const { address, tokenId, hotkey, netuid } = bestBondableBalance
+          handleOpenBittensorModal({
+            address,
+            tokenId,
+            poolId: hotkey as string, // TODO fix typing issue on handler, in practice it's undefined by default
+            netuid,
+            stakeType,
+            isSelectStakeDrawerOpen: !stakeType,
+            step: stakeType !== "subnet" ? "form" : "subnet-form",
```
0xKheops, line 70: @UrbanWill I had to remove the default value of stakeType to allow detecting if user is already staking regardless of the stake type. it does not default to root anymore.
This is why the flags here are now computed differently.
This will conflict with #2193, please be careful when merging.

UrbanWill: Cool, lets get this in first and I'll handle the conflicts there

UrbanWill (commented):
LGTM

UrbanWill (approved):
LGTM

PR #2198, opened 2025-10-10
feat: Yield integration for DeFi

aabid-majid:
New `Earn` tab that replaces the `Staking` tab.
For now, the Earn tab will only have one data source: yield.xyz API.
However it is planned to host earn opportunities/positions for more data source, such as SEEK staking in the near future.
Therefore a clear separation between `Earn***` and `Yieldxyz***` components has been enforced.

Right now we are only scratching the surface when using yield.xyz capabilities, because its results are automatically trimmed down based on which products we have enabled in our yield.xyz org, and we did not enabled many.
No substrate/solana products are enabled, though they might in the future.
No products that require manual claiming, or that have cooldowns, are enabled. though they might in the future.
=> solution may appear a bit overengineered because of that, it has actually been tested from a dev org that has more capabilities. 

Misc improvements:
- new `WizardModalDialog` and `PopupSizeModalContainer` components that should be usable for all our wizards
- new `createGlobalOpenClose` utility that generates a `useGlobalOpenClose` that supports typed args


apps/extension/src/inject/solana/util.ts
```diff
@@ -54,3 +54,49 @@ export const deserializeTransaction = (transaction: string): Transaction | Versi
     return Transaction.from(bytes)
   }
 }
+
+export const deserializeTransactionFromBase64 = (
```
0xKheops: Please dont add code in the `extension/src/inject` folder: this folder is for code that is injected in all websites by the extension, to allow dapps to communicate with the wallet. it needs to be as tiny as possible.

looks like this should go in `packages/solana` instead

apps/extension/src/ui/apps/dashboard/layout/DashboardLayout.tsx
```diff
@@ -71,10 +71,11 @@ const RESPONSIVE_FLEX_SPACING = classNames(
 const NavButton: FC<{
   label: ReactNode
   icon: FC<{ className?: string }>
+  activeIcon?: FC<{ className?: string }>
```
0xKheops: looks unused

apps/extension/src/ui/apps/popup/pages/Earn/DepositConfirm.tsx
```diff
@@ -0,0 +1,194 @@
...
+        txMetadata: (transaction.txDetails as { shortMetadata?: Uint8Array | `0x${string}` })
+          ?.shortMetadata,
+      }
+    } else if (platform === "solana") {
+      return {
+        platform: "solana" as const,
+        networkId: token.networkId as `0x${string}`,
+        payload: transaction.tx as
+          | import("@solana/web3.js").Transaction
+          | import("@solana/web3.js").VersionedTransaction,
```
0xKheops: is there a reason why there are inline imports here?

apps/extension/src/ui/apps/popup/pages/Earn/DepositConfirm.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const [searchParams] = useSearchParams()
+  const navigate = useNavigate()
+  const [currentStep, setCurrentStep] = useState<"confirm" | "execute" | "progress">("confirm")
+  const [transactionStep, setTransactionStep] = useState<1 | 2>(1)
+  const { set, resetUserInput } = useDepositWizard()
+  const { token } = useDepositFunds()
+  const [progress, setProgress] = useState<{ networkId: string; txId: string } | null>(null)
+
+  // Get parameters from URL
+  const account = searchParams.get("account") || ""
```
0xKheops: a bit worried about a modal playing with the page's url, especially with generic names such as `tokenId` and `account` which could be used by the portfolio page itself in the future (or even now, would need to check)
is it possible to keep the wizard state variables in a react context or an observable that would be local to the modal ? if not, please prefix all url params with `earn`, such as `earnTokenId` and `earnAccount`

I imagine this was inspired by the Send Funds wizard, but the Send Funds wizard lives in its own popup page (also using url params there allow to open the page with prefilled parameters)

0xKheops (changes requested):
Would it be possible to rename "yield" to "earn" ?
it's very likely that the earn tab will have more providers (yieldxyz being just the first one), so it would be preferable to have consistency in the naming of what s related to the earn feature.
at first glance main ones to rename are: 
- extension-core/domains/yield folder package
- extension/ui/api methods and associated backend handlers
- hooks names

0xKheops (approved):

PR #2221, opened 2025-10-22
chore: run ff build script without sudo when BUILD_NOSUDO=true

alecdwm:
Changesets is going to complain about missing one, but I think we can ignore it - I only made a small change to `tsconfig.json` to remove an error from vscode


0xKheops (approved):
LGTM!

PR #2227, opened 2025-10-30
fix: clear rpc cache when a network is modified

0xKheops:
makes it so the new RPC order is applied instantly after editing a network


alecdwm (approved):
LGTM

PR #2246, opened 2025-11-10
fix: disappearing balances if fetch error

0xKheops:
fixes an issue occuring on balance modules that poll/fetch balances instead of subscribing (erc20, hydration, bittensor dtao, etc)

in case the RPC call that fetches balances fails, the balance provider will now return the previous balances from storage, marked as stale, instead of considering there is no more balance.

=> balances shouldn't randomly disappear from portfolio anymore


alecdwm (approved):
<div align="center">

excellent

</div>

PR #2247, opened 2025-11-10
feat: deprecate azns on chain ids

0xKheops:
- deprecates azns in `@talismn/on-chain-ids`
- updated `@talismn/on-chain-ids` dependencies to be minimal (removed @polkadot/*, @azns/*, and @talismn/balances)


alecdwm (approved):
Awesome 😎

PR #2249, opened 2025-11-10
Feat/bittensor upgrades

UrbanWill:
## Notes:

- On hold while we figure out the best UX for this feature.

## Description 

Added Claim Settings for Bittensor Root staking.

Added Pending Root Claim Balances to balances lib.

Root staking (subnet 0) rewards are accumulated as alpha across subnets. Every ~2 days an auto claim will be triggered and it'll go over each subnet users accrued those root staking rewards and will claim them.

Now there are two claim types for when the auto claimer claim root staking rewards:

Swap (default): The auto claimer automatically swaps the earned rewards and adds it to the user's current root stake position.

Keep: The auto claimer keeps the user's earned emission in alpha in every subnet. It remains as staked alpha.

------

Nothing changes If a user is staking directly on a subnet that is not root

### 🖼️ **Screenshots:**

https://github.com/user-attachments/assets/7f9e118a-b7d0-4362-9faf-ed0fd88623e0


apps/extension/src/ui/domains/Portfolio/AssetDetails/BittensorClaimSettingsToolbarButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    return {
+      networkId: token.networkId,
+      address: balance.address,
+      netuid: token.netuid,
+      hotkey: token.hotkey,
+      stakeDirection: "bond",
+      step: "claim-settings",
+    }
+  }, [accounts, balances.each, bittensorNetworkIds])
```
0xKheops: ⚠️ Should not use `balances.each` in hook deps, as it creates an array each time it's called, causing the hook to re-evaluate on each render. 
The dependency should just be `balances`

packages/balances/src/modules/substrate-dtao/fetchBalances.ts
```diff
@@ -180,3 +249,35 @@ export const fetchBalances: IBalanceModule<typeof MODULE_TYPE>["fetchBalances"]
...
+  networkId: string,
+  metadataRpc: `0x${string}`,
+  hotkeys: string[],
+): Promise<Map<string, Map<number, bigint>>> => {
+  if (!hotkeys.length) return new Map<string, Map<number, bigint>>()
+
+  const entries = await Promise.all(
+    hotkeys.map(async (hotkey) => {
+      try {
+        const rootClaimable = await fetchStorageValue<[number, bigint][] | null>(
```
0xKheops: all fetchs could be merged in a single query to reduce the risk of hitting rpc rate limits

packages/balances/src/modules/substrate-dtao/calculatePendingRootClaimable.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    pendingRootClaimBalances.push({
+      address,
+      tokenId: subDTaoTokenId(networkId, netuid, hotkey),
+      baseTokenId: subDTaoTokenId(networkId, netuid),
+      hotkey: hotkey,
+      netuid: netuid,
+      scaledAlphaPrice,
+      pendingRootClaim,
+      stake: pendingRootClaim,
```
0xKheops: this makes the pending root claim amount end up in the both in the `free` and `locked` properties of the balance, one of the symptoms is causing unstake max (with max button) transactions to fail.
i think it should be 0n, but maybe i missed something?

apps/extension/src/ui/domains/Portfolio/AssetDetails/BittensorUnstakeButton.tsx
```diff
@@ -33,7 +33,9 @@ export const BittensorUnstakeButton: FC<{ balances: Balances; className?: string
 
     const token = balance?.token as SubDTaoToken
 
-    return balance && token
+    const hasFreeBalance = balance?.free.planck && balance?.free.planck > balance?.locked.planck
```
0xKheops: see my comment in `fetchBalances.ts`, once fixed this could be changed to `!!balance.free.planck`.

apps/extension/src/ui/domains/Portfolio/AssetDetails/TokenContextMenu.tsx
```diff
@@ -94,6 +95,7 @@ const StakeMenuItem: FC<{ tokenId: string }> = ({ tokenId }) => {
 
 type Props = {
   tokenId: TokenId
+  balances: Balances
```
0xKheops: seems unused

apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/index.tsx
```diff
@@ -24,5 +25,7 @@ export const BittensorBondModalRouter = () => {
       return stakeType === "subnet" ? <BittensorSubnetBondReview /> : <BittensorRootBondReview />
     case "follow-up":
       return <BittensorBondFollowUp />
+    case "claim-settings":
```
0xKheops: see above comment

apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorClaimSettings.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { useBittensorBondModal } from "../../hooks/useBittensorBondModal"
+import { useBittensorBondWizard } from "../../hooks/useBittensorBondWizard"
+import { BittensorStakingModalHeader } from "../BittensorModalHeader"
+import { BittensorModalLayout } from "../BittensorModalLayout"
+import { BittensorAssetAccountSummary } from "../components/BittensorAssetAccountSummary"
+
+export const BittensorClaimSettings = () => {
+  const [selectedClaimType, setSelectedClaimType] = useState<RootClaimType>(DEFAULT_ROOT_CLAIM_TYPE)
+  const { t } = useTranslation()
+  const { nativeToken, account, accountPicker, setAddress, onSubmitted } = useBittensorBondWizard()
```
0xKheops: feels inappropriate to reuse the bond wizard hook for this unless it's actually part of it. otherwise this form should live in its own modal.

packages/balances/src/modules/substrate-dtao/getMiniMetadata.ts
```diff
@@ -48,7 +48,13 @@ const getData = (metadataRpc: `0x${string}`): `0x${string}` | null => {
...
+    [
+      {
+        pallet: "SubtensorModule",
+        items: ["RootClaimable"],
+      },
+      { pallet: "SubtensorModule", items: ["TransferToggle"] },
+    ],
     [
       {
         runtimeApi: "StakeInfoRuntimeApi",
```
0xKheops, line 54: compactMetadata expects only one entry per pallet, with all the required items.
if changed to this, the issue we talked about will go away: `[{ pallet: "SubtensorModule", items: ["TransferToggle", "RootClaimable"] }],`

you may need to click the Reload button in Talisman Settings to clear the db and have it regenerated.

please dont forget to cleanup the fallback metadata download in `fetchStorageValues` afterwards

0xKheops (changes requested):
Great one over all, just easy to fix stuff

However It feels like the claim setting form could also be integrated to the root staking flow, with the setRootClainType call being bundled to the tx if user desires to change it.
the button to open the step could be added to the root staking bond review form.
it would both help users to understand what's going on, and simply notice the feature, in context.
it would also justify coupling the form with the useBittensorBondWizard hook :)

packages/balances/src/modules/substrate-dtao/getMiniMetadata.ts
UrbanWill: Thanks for the tip!

However, after adding the suggested changes it still doesn't work. I am leaving the fall back until this issue has been resolved.

I suspect it has something to do with the new metadata v8 changes required for this PR, can you please have a look? The reason for the suspicion is the following error "githubChaindata.ts:40 Failed to fetch chaindata Error: Failed to fetch from https://cdn.statically.io/gh/TalismanSociety/chaindata/main/pub/v8/chaindata.min.json"

It would be great to have the metada updating process documented, so I can reliably update it and reproduce any issues.

0xKheops: Reason you are still having issue is that chaindata provider is still targeting the `main` branch of chaindata (which uses v7).
it needs to target the branch that was bumped to v8, like so in `chaindata-provider/src/constants.ts`
Screenshot 2025-12-20 at 11 53 37 AM

Works fine for on my machine afterwards.

Yeah i hear you for the doc, however it's a big one that i can't do on the top of my head, it has to be done while actually going through the whole process.
So i planned to add it while implementing Solana Token 2022 balance module, which should be relatively soon.

Let's wrap this PR on monday, can do a screen sharing session if there is still any issue.

apps/extension/src/ui/domains/Staking/Bittensor/utils/constants.ts
```diff
@@ -10,3 +12,5 @@ export const TALISMAN_FEE_RECEIVER_ADDRESS_BITTENSOR =
   "5DzsVV2L4M9r4uWoyarzPyhfeCv6DDAEs5rM2bpHjmerPcGa"
 
 export const TALISMAN_FEE_BITTENSOR = 0.3
+
+export const DEFAULT_ROOT_CLAIM_TYPE: RootClaimType = "Keep"
```
0xKheops: The on-chain default is still `Swap`, eventhough they are planning to change it to `Keep` soon

```suggestion
export const DEFAULT_ROOT_CLAIM_TYPE: RootClaimType = "Swap"
```

packages/balances/src/modules/substrate-dtao/calculatePendingRootClaimable.ts
```diff
@@ -0,0 +1,58 @@
...
+    // We'll need to handle the fixed-point multiplication correctly
+    // For now, we'll do a simple multiplication and assume I96F32 is represented
+    // as a bigint with appropriate scaling
+
+    // Multiply claimable_rate by root_stake
+    // I96F32 multiplication: (a * b) / 2^32 to handle fixed-point
+    // But without knowing the exact scaling, we'll do simple multiplication
+    // This might need adjustment based on actual I96F32 implementation
+    const pendingRootClaim = (stake * claimableRate) / 2n ** 32n
+
```
0xKheops, line 39: My second brain is suggesting this, to match the python SDK implementation.
It's saying current code is slightly undercounting, though I didn't observe any difference in outputs after testing the change.
So up to you, this might be the fix to the issue mentionned in your comment there?

```suggestion

    // Multiply claimable_rate by root_stake
    // I96F32 multiplication: round((a * b) / 2^32)
    const pendingRootClaim = (stake * claimableRate + (1n << 31n)) >> 32n

```

0xKheops (changes requested):
the wizard looks great!
just a default value to fix and we're good

packages/balances/src/modules/substrate-dtao/calculatePendingRootClaimable.ts
UrbanWill: Oh that comment was left when I wasn't entirely sure how to work with the I96F32 and should have been removed by now.

HOWEEVER your second brain is right, I've cross reference the usage of I96F32 in bittensor's repo and they round it to to the nearest number, I was truncating.

0xKheops (approved):
LGTM!

PR #2271, opened 2025-12-02
feat: use signed ramp api

alecdwm:
This PR uses our new proxy to (a) sign the url which opens the ramp widget and (b) insert our ramp api key to all ramp api requests.

These changes are necessary due to some security improvements on the ramp api side.

Serverside code for signing the ramp widget URL is here:
https://github.com/TalismanSociety/ramp-api/blob/main/src/talismanEndpoints.ts#L19-L33

And the proxy code is here:
https://github.com/TalismanSociety/ramp-api/blob/main/src/proxyRampApi.ts#L11-L14


apps/extension/src/ui/domains/Ramps/ramp/helpers.ts
```diff
@@ -30,7 +19,13 @@ export const getRampBuyUrl = async (
     fiatValue: amount.toString(),
   })
 
-  return `${RAMPS_RAMP_PAY_URL}/?${params.toString()}`
+  const response = await fetch(
+    `${RAMPS_RAMP_API_URL}/talisman/getSignedBuySellUrl?${params.toString()}`,
+  )
+  if (!response.ok) throw new Error(`Failed to generate Ramp URL`)
+
+  const url = (await response.json())?.url
```
0xKheops, line 27: maybe throw if `!url` ?

0xKheops (dismissed):
LGTM! just one comment, up to you if you want to address

PR #2281, opened 2025-12-23
fix: add APY display to subnet review and change staking text to badge

themorlock:


themorlock:
Added APY to the review pages for Root staking and Subnet staking.

Also removed the "You are Staking/Unstaking" text from the subnet review page to fit the APY. The root staking page had enough room for the APY but I still also the "You are Staking/Unstaking" text from it to be consistent.

dtao_stake

dtao_unstake

root_stake

root_unstake

0xKheops (approved):
LGTM!

PR #2282, opened 2025-12-26
fix: update min dtao staking amount to account for fees

themorlock:
first_time_stake
second_time_stake

When staking, the minimum amounts now take into account the swap fee + the Talisman fee (0.3% but can change based on SEEK ownership).

Unstaking didn't seem to need a fix because the way we collect fees when unstaking is we unstake the amount the user types in, then remove our cut from the converted TAO. This is unlike staking where we remove our fee from the amount they type in.

The 2 images have different minimums because when you first stake on a subnet, you need at least `NominatorMinRequiredStake`, but for subsequent stake actions you need to add at least `DefaultMinStake`.


apps/extension/src/ui/domains/Staking/Bittensor/hooks/useBittensorStakingPayload.ts
```diff
@@ -69,6 +75,38 @@ export const useBittensorStakingPayload = ({
     return taoToAlpha(minTaoStake, alphaPrice)
   }, [minTaoStake, alphaPrice])
 
+  // simulate swap for minimum stake to get swap fee
+  const { data: minStakeSwapSimulation } = useQuery({
+    queryKey: ["minStakeSwapSimulation", sapi, netuid, minTaoStake?.toString()],
```
0xKheops: sapi object wont serialize consistently, we have a id property to workaround that

```suggestion
    queryKey: ["minStakeSwapSimulation", sapi?.id, netuid, minTaoStake?.toString()],
```

0xKheops (changes requested):
Looks great, appreciate the tests!
Just a lil change needed for the useQuery key

0xKheops (approved):

PR #2284, opened 2025-12-29
fix: remove discount tiers from default remote config

UrbanWill:
# Description

- Fixes bug where 3% discount was being applied instead of 15%.

### 🗒️ **Notes:**

This issue only occurs for users staking 10k+ SEEK tokens.

The issue was caused because 'real' remove config values get merged with `DEFAULT_REMOTE_CONFIG` values, and I was under the assumption that the correct remote config would replace the default values.

I ended up writing quite a few tests to figure out what as going on and went the wrong way about it, it was not a bad math issue, simply loading the incorrect default values.


0xKheops (approved):
LGTM

PR #2285, opened 2025-12-29
fix: added storage external addresses vs token

themorlock:
Added a store of confirmed addresses so users can click once and not be reminded again.

Screenshot 2025-12-29 at 4 13 58 PM
Screenshot 2025-12-29 at 4 16 01 PM


0xKheops (changes requested):
Looks great, however i made a mistake when describing what to do.
blob stores are only for temporary data, as they are all flushed when clicking the "reload" button in Talisman settings page.
Can you replace the blob store with a store that uses StorageProvider ? (same as what we use for authorisedSites, settings, app, keyring, etc)

0xKheops (approved):
LGTM

PR #2287, opened 2026-01-01
fix: fixed long network names blocking buttons

themorlock:
Fixed bug where long network names would stop user from clicking action buttons.

Screenshot 2026-01-01 at 3 58 52 PM


0xKheops (approved):
LGTM

PR #2288, opened 2026-01-01
fix: added details in history for Bittensor dTAO operations

themorlock:
Added TAO/Alpha information for dTAO staking events (unstake/stake).
Screenshot 2026-01-01 at 4 02 05 PM

It seems like even for swap details (SimpleSwap), we still just say "Confirmed/Failed", so I didn't change the dTAO history items to stay "Stake/Unstake" (but this can easily be added if we want).

### DIFF
--- a/apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorRootBondReview.tsx
+++ b/apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorRootBondReview.tsx
@@ -40,14 +40,12 @@ export const BittensorRootBondReview = () => {
     // Root staking: TAO -> TAO (no alpha token conversion)
     return {
       type: "bittensor-staking",
-      action: stakeDirection === "bond" ? "stake" : "unstake",
       fromTokenId: nativeToken.id,
       toTokenId: nativeToken.id,
       fromAmount: amountIn.toString(),
       toAmount: amountIn.toString(), // same amount for root staking
-      netuid: 0, // root staking
     }
-  }, [nativeToken?.id, amountIn, stakeDirection])
+  }, [nativeToken?.id, amountIn])
 
   useEffect(() => {
     // enable confirm button 0.5 second after the screen is open, to ensure the user doesnt accidentally click it (ex: double click from prev screen)
--- a/apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorRootBondReview.tsx
+++ b/apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorRootBondReview.tsx
@@ -1,3 +1,4 @@
+import { subDTaoTokenId } from "@talismn/chaindata-provider"
 import { InfoIcon } from "@talismn/icons"
 import { WalletTransactionInfo } from "extension-core"
 import { useEffect, useMemo, useState } from "react"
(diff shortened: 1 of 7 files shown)


themorlock:
Consolidated stake and unstake into a single type. Also added TAO <-> ALPHA in history.

Screenshot 2026-01-06 at 1 29 29 PM
Screenshot 2026-01-06 at 1 44 18 PM

packages/extension-core/src/domains/transactions/types.ts
```diff
@@ -58,6 +58,15 @@ export type WalletTransactionInfo =
...
     }
+  | {
+      type: "bittensor-staking"
+      action: "stake" | "unstake"
+      fromTokenId: TokenId
+      toTokenId: TokenId
+      fromAmount: string
+      toAmount: string
+      netuid: number
+    }
```
0xKheops, line 67: I think both `action` and `netuid` should be removed as they can be derived from token ids:
- `netuid` seems unused, and unnecessary anyways.
- to determine if it's a stake or unstake, can use `parseTokenId(txInfo.fromTokenId).type === "substrate-native"`

packages/extension-core/src/domains/transactions/helpers.ts
```diff
@@ -270,3 +270,6 @@ export const isTxInfoTransfer = (txInfo: WalletTransactionInfo | undefined | nul
 
 export const isTxInfoApproval = (txInfo: WalletTransactionInfo | undefined | null) =>
   isTxInfoOfType(txInfo, "approve-erc20")
+
+export const isTxInfoBittensorStaking = (txInfo: WalletTransactionInfo | undefined | null) =>
+  isTxInfoOfType(txInfo, "bittensor-staking")
```
0xKheops: it feels like `bittensor-staking` should be one of the types of the `isTxInfoSwap` helper, as we want to present it as a swap.
it could simplify the activity list code

0xKheops (changes requested):
looks great now, but can be further simplified

apps/extension/src/ui/domains/Staking/Bittensor/BittensorBondModal/Forms/BittensorRootBondReview.tsx
```diff
@@ -34,6 +35,18 @@ export const BittensorRootBondReview = () => {
 
   const [isDisabled, setIsDisabled] = useState(true)
 
+  const txInfo: WalletTransactionInfo | undefined = useMemo(() => {
+    if (!nativeToken?.id || amountIn === null) return undefined
+    // Root staking: TAO -> TAO (no alpha token conversion)
+    return {
+      type: "bittensor-staking",
+      fromTokenId: nativeToken.id,
+      toTokenId: nativeToken.id,
```
0xKheops: if using native token id for both, the label in the details form will be incorrect for unstake txs.
one of them is not the native token but alpha for netuid 0, can generate its id using subDTaoTokenId("bittensor", 0).

apps/extension/src/ui/domains/Transactions/TxHistory/TxHistoryDetails/TxHistoryDetailsTxInfo.tsx
```diff
@@ -174,3 +176,44 @@ const SwapLifiTxInfo: FC<{
...
+
+  return (
+    <TxInfoCard>
+      <Trans
+        t={t}
+        defaults={
+          isStake
+            ? "Stake <FromTokens /> for <ToTokens />"
+            : "Unstake <FromTokens /> for <ToTokens />"
+        }
```
0xKheops: i18n strings cannot be conditional or they wont be parsed and sent to our translation platform.
sadly we need 2 different `<Trans />` components here.

0xKheops (changes requested):
2 little issues to fix

apps/extension/src/ui/domains/Transactions/TxHistory/TxHistoryDetails/TxHistoryDetailsTxInfo.tsx
0xKheops: or maybe have the operation name being output by a 3rd entry in the components property

0xKheops (approved):

PR #2289, opened 2026-01-04
fix: added functionality to select subnets to keep emissions on

themorlock:
Added option to select subnets to keep emissions on.

Screenshot 2026-01-04 at 3 52 46 PM

Screenshot 2026-01-04 at 3 52 54 PM


0xKheops (changes requested):
Please few adjustments before merging:
- move the useEffect() that sets claim mode + subnets from forms to the wizard hook, so that it executes only once when the modal is open (atm, it s reset each time we switch form, which feels wrong when using the back button from the subnets form)
- have the wizard hook compare on-chain value with in memory value and output a `canSubmit` flag that is used to have the submit buttons enabled or not. this is to prevent submitting unnecessary transactions.
- remove alpha in pool and tao in pool sort options

0xKheops (approved):

PR #2295, opened 2026-01-09
feat: Christmas tooling update 🎅🏼

0xKheops:
This PR replaces our test/build pipeline:
- Before: preconstruct, turbo, webpack (+x plugins), jest, prettier (+ x plugins), eslint (+ x plugins)
- After: tsup, wxt, vitest, biome

Important: 
- WXT dev mode does not support firefox MV3 yet, so MV3 is only used for prod builds, while MV2 is used for firefox dev builds.
- Biome scripts have the --changed flag to process only the flles that are changed compared to dev branch. this is to avoid a big bang PR where all files would be reformatted. this can be done later when there are no other pending PRs in the repo

⚠️ Because we need ES Modules support:
- new minimum chrome version: 121 (January 23, 2024)
- new minimum firefox version: 128 (July 9, 2024)

Note: hot reloading (rebuilding) the background script is pretty long for now, like 10+ seconds. it can be improved in another "big bang refactor" PR after this one is merged. it will require separating code that is background-only and code that needs to be shared by frontend and background, in 2 separate packages, so wxt can treeshake background script more efficiently.

Misc: 
- installed [knip](https://knip.dev/) and removed 35 unused dependencies and 40 unused files with it. we will be able to integrate knip checks in the CI if we wish to, in the future.
- updated CI to run on node 24 instead of 18
- updated caniuse list which was 15 months old
- added a global `pnpm typecheck` script, which is now included in CI
- added a global `pnpm check` that performs biome checks on files that are different from target branch
- packages dont need to be built anymore, except when publishing them


UrbanWill (changes requested):
Ho Ho Ho Talisman engineering xmas is finally here! 🎅 

I am very happy to see these changes going in, code looks good and everything works as expected.

Only one minor thing, import warnings. Eg: Libs like `"@talismn/util"` or `"@talismn/icons"` and some type imports here and there. Perhaps a biome config missing?

UrbanWill (dismissed):
Previous issues fixed by running `pnpm build:packages`

PR #2296, opened 2026-01-10
Feat/change subnet validator

themorlock:
Added feature to change validator

Screenshot 2026-01-09 at 7 05 42 PM
Screenshot 2026-01-09 at 7 05 55 PM


apps/extension/src/ui/domains/Staking/Bittensor/hooks/useBittensorChangeValidatorModal.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { useGlobalOpenClose } from "@talisman/hooks/useGlobalOpenClose"
+
+import type { ChangeValidatorOpenOptions } from "./useBittensorChangeValidatorWizard"
+import { useResetBittensorChangeValidatorWizard } from "./useBittensorChangeValidatorWizard"
+
+export const useBittensorChangeValidatorModal = () => {
+  const reset = useResetBittensorChangeValidatorWizard()
+
+  const { isOpen, open: innerOpen, close } = useGlobalOpenClose("BittensorChangeValidatorModal")
```
0xKheops: please use createGlobalOpenClose instead of useGlobalOpenClose + a reset hook.
it will allow you to pass an argument, can pass the tokenId there, so you can filter the positions (in position picker) based on it.
some examples available in yieldxyz modals

0xKheops (approved):

PR #2309, opened 2026-01-25
fix: json import for polkadot.js accounts (scalar validation error)

xx-wallet-dev:
# Fix JSON Import for Polkadot.js Accounts

## Problem
Importing Polkadot.js JSON account backups into Talisman failed with: expected valid scalar error for sr25519/ed25519 accounts.

## Root Cause
Secret key format mismatch between Polkadot.js PKCS8 format ([seed + publicKey]) and Talisman's internal format ([privateKey + nonce]).

## Solution
Implemented format conversion by extracting seed from Polkadot.js JSON and deriving proper secretKey using Talisman's deriveKeypair function (same approach as addAccountDerive for mnemonic accounts).

## Changes
- apps/extension/src/ui/domains/Account/AccountAdd/AccountAddJson/context.ts: Added state preservation, JSON array format support, format conversion in importAccounts
- packages/extension-core/src/domains/keyring/getSecretKeyFromPjsJson.ts: Now returns {secretKey, publicKey} object
- packages/extension-core/src/domains/keyring/migrations/migrateFromPjsKeyring.ts: Updated to use .secretKey property

## Testing
- Tested with multiple Polkadot.js JSON exports (sr25519 and ed25519)
- Tested with both single JSON files and array format
- All imports succeed without scalar validation errors

## Impact
- Enables seamless migration from Polkadot.js Extension to Talisman
- Supports both legacy and modern JSON export formats
- No breaking changes

### DIFF
--- a/apps/extension/src/ui/domains/Account/AccountAdd/AccountAddJson/context.ts
+++ b/apps/extension/src/ui/domains/Account/AccountAdd/AccountAddJson/context.ts
@@ -2,15 +2,22 @@ import { createPair } from "@polkadot/keyring"
 import type { KeyringPair, KeyringPair$Json } from "@polkadot/keyring/types"
 import type { KeyringPairs$Json } from "@polkadot/ui-keyring/types"
 import { assert, hexToU8a, isHex, u8aToString } from "@polkadot/util"
-import { base64Decode, decodeAddress, encodeAddress, jsonDecrypt } from "@polkadot/util-crypto"
+import {
+  base64Decode,
+  base64Encode,
+  decodeAddress,
+  encodeAddress,
+  jsonDecrypt,
+} from "@polkadot/util-crypto"
 import type { EncryptedJson, KeypairType } from "@polkadot/util-crypto/types"
 import { provideContext } from "@talisman/util/provideContext"
 import { type Address, Balances } from "@talismn/balances"
-import { encodeAnyAddress, isAddressEqual, normalizeAddress } from "@talismn/crypto"
+import { deriveKeypair, encodeAnyAddress, isAddressEqual, normalizeAddress } from "@talismn/crypto"
 import { api } from "@ui/api"
 import { useAccountImportBalances } from "@ui/hooks/useAccountImportBalances"
 import { useAccounts, useNetworks } from "@ui/state"
 import type { Account, LegacyAccountOrigin } from "extension-core"
+import { getSecretKeyFromPjsJson } from "extension-core/domains/keyring/getSecretKeyFromPjsJson"
 import { log } from "extension-shared"
 import { useCallback, useEffect, useMemo, useState } from "react"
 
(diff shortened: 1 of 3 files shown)


0xKheops:
Hey @xx-wallet-dev 
Thank you for this PR.

Before we look into it, could you please provide us with exact steps to create a json file that can be used to reproduce the the scalar validation error ?
i ve never experienced it and this code has been there for a long time, so i'm quite surprised.

xx-wallet-dev:
Hi, apologies for the delayed response. I was offline for a while and couldn't reply sooner ;-):

### Steps to Reproduce the Scalar Validation Error

1. **Export an account from the XX Network Wallet** (https://wallet.xx.network)
   - Create or use an existing SR25519 account
   - Export the account as JSON file (with password)

2. **Try to import this JSON file into Talisman**
   - Go to "Add Account" → "Import from JSON"
   - Select the exported JSON file
   - Enter the correct password

3. **Result:** "Invalid scalar" or "expected valid scalar" error

### Root Cause Analysis

I encountered a similar issue when building a Go-based wallet generator for the XX Network Sleeve Wallet. The core problem is that **Polkadot.js uses a proprietary PKCS8 format** (not standard RFC 5958).

#### Proprietary PKCS8 Structure (117 bytes unencrypted)
Offset | Size | Field
-------|------|------------------
0 | 16 | PKCS8_HEADER [48, 83, 2, 1, 1, 48, 5, 6, 3, 43, 101, 112, 4, 34, 4, 32]
16 | 64 | private_key (Ed25519 expanded key)
80 | 5 | PKCS8_DIVIDER [161, 35, 3, 33, 0]
85 | 32 | public_key ← The correct public key is HERE!

```
// The Problem in Talisman's `createPairFromJson()`
typescript
return createPair(
{ toSS58: encodeAddress, type: cryptoType },
{ publicKey: decodeAddress(address, true) }, // ← Uses address field!
meta,
decoded,
encType
)
```


Talisman extracts the public key from the `address` field via `decodeAddress()`, but **Polkadot.js's `decodePair()` extracts it from the encrypted PKCS8 payload** (bytes 85-117):

```
// From @polkadot/keyring - packages/keyring/src/pair/decode.ts
const pubOffset = divOffset + PAIR_DIV.length;
const publicKey = decrypted.subarray(pubOffset, pubOffset + PUB_LENGTH);
```

When the JSON is exported from a wallet that uses a different SS58 prefix (like XX Network with prefix 55), the public key derived from decodeAddress(address) may not match the actual public key stored inside the encrypted payload. SR25519 validation then fails with "expected valid scalar" because the keypair doesn't match.
Why This Code "Worked Before"
This code path likely worked for standard cases because:
Accounts created within Talisman or Polkadot.js always have the address derived from the same public key
Cross-wallet imports with different SS58 encodings weren't commonly tested
With XX Network wallets (SS58 prefix 55), the mismatch becomes apparent.
Suggested Fix
Extract the public key from the decrypted PKCS8 data (bytes 85-117) instead of from the address field, matching what Polkadot.js does in decodePair().
Let me know if you need a sample JSON file for testing. I can provide one (with encrypted test data, not real private keys).

prury:
@xx-wallet-dev i've followed your steps and was able to import a JSON file that i've exported directly from https://wallet.xx.network/ into Talisman: 

image

the only error i got was when trying to import the batch file: 

image

xx-wallet-dev:
Thanks @prury for taking the time to test this, I really appreciate it!
That's very interesting feedback. We have wallets generated with different tools (Go-based generator, Python substrateinterface, and direct exports from wallet.xx.network) and so far haven't been able to pinpoint which exact combination triggers the scalar error. Your finding that the standard export/import works is valuable input.
I'll investigate this more thoroughly with our different wallet files and get back to you once we have a concrete, reproducible case. Feel free to close this PR in the meantime.

xx-wallet-dev:
Thanks @prury for testing and for the valuable feedback. It pointed me in the right direction.

After further investigation, I found the actual issue: it's in our own wallet generation tool (a Go-based JSON export), not in Talisman. I were building the 64-byte secret key in the PKCS8 payload incorrectly (`seed + seed` instead of a proper SR25519 expanded secret key with scalar + nonce seed). Talisman's code is correct.
Apologies for the noise. Closing this PR now. Thanks again for your time.

PR #2316, opened 2026-02-05
test: change passethub settings

prury:
Since the last chaindata change and some changes on Paseo testnet, passet hub is now part of Paseo Asset Hub. This PR changes the tesnet network/token used in the testnet transfer e2e test.


0xKheops (approved):
LGTM!

PR #2321, opened 2026-02-11
fix: "Optimism: Improve Error message" bug

themorlock:
Fixed "Optimism: Improve Error message" bug. Now shows "Insufficient ETH" if ETH balance is too low.
Screenshot 2026-02-11 at 12 58 14 AM


apps/extension/src/ui/domains/Ethereum/useIsValidEthTransaction.ts
```diff
@@ -42,9 +42,15 @@ export const useIsValidEthTransaction = (
       // balance checks
```
0xKheops, line 42: looks like we could use a `if(typeof balance !== "bigint") throw new Error(t("Failed to load ${nativeSymbol} balance"))` as a first check

themorlock: Makes sense. Added it

0xKheops (approved):
LGTM!

prury (approved):

PR #2323, opened 2026-02-13
fix: "Change Validator" doesn't show up if there isn't a validator to…

themorlock:
Fixed "Change validator bug shouldnt show if no position" bug. "Change Validator" button only shows up now if there is a validator to change.
Screenshot 2026-02-13 at 12 20 54 AM


0xKheops (approved):
LGTM!

should be good this time but in the future please always use isAddressEqual when matching substrate addresses, just in case one of them isn't encoded with 42 prefix

prury (approved):

PR #2331, opened 2026-02-25
feat: improve paraverse protector

0xKheops:
⚠️ This PR bumps database schema

- fetch lists from cdn instead of github api, eliminating rate limit issues
- store lists in blob storage, which saves ~5mb and simplifies db schema
- update from source repos every 5 minutes
- fix mm file location
- refactored paraverse protector to not use a singleton so it can be tested and disposed properly


themorlock (approved):
Looks great!

prury (approved):

PR #2365, opened 2026-03-23
feat: expired quote detection

0xKheops:
with LIFI, if a tx fails to submit because quote isnt valid anymore, it should open a drawer to inform the user that quote needs to be refreshed. when closing the drawer, quote will refresh.


0xKheops:
current status: cant reproduce the issue to test the fix

PR #2396, opened 2026-04-17
feat(chaindata-provider): support optional chaindataSlimUrl override

UrbanWill:
Allow ChaindataProvider consumers to pass a custom that replaces the default consolidated chaindata URL when fetching.

- Add  to ChaindataProviderOptions
- Thread the URL through getDefaultChaindata$ → getGithubChaindata$ → fetchChaindata
- Convert githubChaindata$ singleton into getGithubChaindata$ factory
- Remove stray console.log from net.ts

### DIFF
--- a/.changeset/quick-cars-pull.md
+++ b/.changeset/quick-cars-pull.md
@@ -1,0 +1,5 @@
+---
+"@talismn/chaindata-provider": minor
+---
+
+added optional chaindata slim url arg
--- a/packages/chaindata-provider/src/provider/ChaindataProvider.test.ts
+++ b/packages/chaindata-provider/src/provider/ChaindataProvider.test.ts
@@ -19,7 +19,7 @@ import { ChaindataProvider, type ChaindataStorage } from "./ChaindataProvider"
 let githubSubject: Subject<Chaindata>
 
 vi.mock("../state/githubChaindata", () => ({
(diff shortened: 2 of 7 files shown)


0xKheops:
imo the url shouldn't be left to the consumer to decide, one of the benefits of current implementation is the version folder being tied to the version of the schemas defined in the library (ex `pub/v9`). If consumer were to set the url, then they couldn't bump the version of the chaindata-provider library without verifying which version of schemas the new version needs, and adjust the url accordingly every time. 
=> it would introduce maintenance steps and risk of errors. 

would an optional `mode: "full" | "slim"` argument on `ChaindataProviderOptions` work instead? Then it would keep existing url crafting logic under the hood, but would target either `chaindata.min.json` or `chaindata.slim.min.json` based on that argument.

@UrbanWill  wdyt?

alecdwm:
I merged https://github.com/TalismanSociety/talisman/pull/2412 into this branch so that I can get a snapshot version of the balances packages which includes both sets of changes.

alecdwm:
Ok we merged https://github.com/TalismanSociety/talisman/pull/2412 into dev, so I rebased this branch off of latest dev!

alecdwm:
Closing in favour of https://github.com/TalismanSociety/talisman/pull/2469 !!! 🎉

PR #2397, opened 2026-04-19
feat: substrate asset discovery

0xKheops:
Handles automatic enabling of `isDefault:false` networks for users that have balances on them.

Triggers every 7 days, or after an account is added.
Scans all substrate networks with `isDefault:false` and `isTestnet:false` that have not been explicitedly been disabled. 
If any account is found alive on the network, enables the network


prury (approved):

alecdwm (approved):
LGTM

PR #2399, opened 2026-04-23
fix: dev improvements

alecdwm:
1. The pre-commit hook will now just format any unformatted code instead of complaining about it,
    but it still blocks and complains if there's any lint errors.
3. Added a console warning during development when balance fetching is disabled, so it's easier to notice.


0xKheops (approved):
LGTM!

PR #2405, opened 2026-04-29
E2E & API test suite expansion and local devnet support via docker image(anvil)

prury:
**Summary:**

Expands the Playwright test suite with new E2E and API tests, and adds local devnet infrastructure for Ethereum (Anvil)

### Test additions
- **API health tests**: Comprehensive tests for Bittensor SN45 and TAO data APIs, with a scheduled CI cronjob (every 30 min)
- **E2E tests**: New specs for accounts, passwords, language settings, custom networks, visual regression, and performance
- **Transfer tests**: Refactored to support local Anvil devchain for Ethereum transfers

### Infrastructure
- **`docker-compose.test.yml`**: Local devnet with Anvil (Ethereum, chain ID 1) 
- **CI**: Added Anvil service to CI workflow, API health cronjob (`api-health.yml`)

### Other changes
- Added `data-testid` attributes across dashboard and account components for better test targeting
- Refactored Playwright fixtures with `useDevChains` and `importAccount` helpers
- Separated Playwright projects so `--project=api` and `--project=chromium` can run independently


.github/workflows/ci.yml
```diff
@@ -112,10 +112,40 @@ jobs:
         run: pnpm exec playwright install --with-deps chromium
       - name: Install xvfb
         run: sudo apt-get update && sudo apt-get install -y xvfb
+      - name: Start Anvil
+        run: |
+          docker run -d --name anvil -p 8545:8545 \
+            ghcr.io/foundry-rs/foundry:latest \
```
0xKheops: Using :latest means CI runs an unverified, mutable image.
Please pin to a well known good hash

.github/workflows/ci.yml
```diff
@@ -112,10 +112,40 @@ jobs:
...
+            "anvil --host 0.0.0.0 --chain-id 1 --block-time 1"
+          # Wait for Anvil to be ready
```
0xKheops, line 120: might want to handle the case where anvil fails to start, and output a clear error. 

ex:

```bash
if ! curl -sf -X POST http://localhost:8545 ...; then
   echo "ERROR: Anvil failed to start after 30 seconds"
   exit 1
 fi
```

playwright/api-tests/bittensor-sn45.spec.ts
```diff
@@ -0,0 +1,516 @@
+/// <reference types="node" />
+
+import { expect, test } from "@playwright/test"
+
+const SN45_API_URL = "https://sn45api.talisman.xyz"
+
+const SN45_API_KEY = process.env.SN45_API_KEY
+const authHeaders: Record<string, string> = SN45_API_KEY ? { "X-API-Key": SN45_API_KEY } : {}
+
+const TEST_NETUID = Math.floor(Math.random() * 125) + 1
```
0xKheops: using a different netuid on each run might lead to unexpected results and make it hard to reproduce errors.
it should be hardcoded to a well known one.

playwright/e2e-tests/password.spec.ts
```diff
@@ -1,89 +1,120 @@
...
+    await popupPage.goto(`chrome-extension://${extensionId}/popup.html`)
+    await popupPage.setViewportSize({ width: 400, height: 600 })
 
-      // Extension pages close on lock — open a fresh popup to unlock
-      const loginPage = await context.newPage()
-      await loginPage.goto(`chrome-extension://${extensionId}/popup.html`)
-      await loginPage.setViewportSize({ width: 400, height: 600 })
+    // Open navigation drawer and lock the wallet
+    await popupPage.getByRole("button", { name: "More" }).click()
+    await popupPage.getByText("Lock Wallet").click()
```
0xKheops: Breaks on my machine. here is copilot findings:

"Lock Wallet" click succeeds but the browser/page closes before the click action completes, causing Target page, context or browser has been closed.
The test expects to click "Lock Wallet" on the popup, but locking the wallet closes all extension pages — including the popup — mid-click. The test
does handle this (line 104 opens a fresh page), but the error fires before it gets there.

This is a race condition in the test. It should catch the page-close error, e.g.:
```ts
await popupPage.getByText("Lock Wallet").click().catch(() => {})
```

0xKheops (changes requested):
Please see below for few adjustments.
Also, i'm not sure about testing SN45 and TDA as part of the CI of this project. IMO it should be done in their respective repos instead.  
but if we want to keep this then:
- either make it run less often (currently every 30 mins, could be every 4 hours instead) or reduce the scope (currently it tests all endpoints, which feels unnecessary)  
- make it send a notification to us in case it fails.

.github/workflows/ci.yml
prury: using a fixed image now

.github/workflows/ci.yml
prury: done!

playwright/api-tests/bittensor-sn45.spec.ts
prury: it is now hardcoded

playwright/e2e-tests/password.spec.ts
prury: error now gets handled properly

prury:
> Please see below for few adjustments. Also, i'm not sure about testing SN45 and TDA as part of the CI of this project. IMO it should be done in their respective repos instead. but if we want to keep this then:
> 
>     * either make it run less often (currently every 30 mins, could be every 4 hours instead) or reduce the scope (currently it tests all endpoints, which feels unnecessary)
> 
>     * make it send a notification to us in case it fails.

Thank you for your insights, changed the time to every 4 hours and added a Discord webhook notification when success/fails

0xKheops (approved):
LGTM

PR #2406, opened 2026-05-01
feat: added Keyring methods for mobile app compatibility

alecdwm:
Added 3 self-explanatory methods:

- `Keyring::hasPassword`
- `Keyring::verifyPassword`
- `Keyring::initializePassword`

`hasPassword` so we can determine if a password has already been set in the keyring (to show onboarding or not)
`verifyPassword` so we can test if an entered password is correct (when unlocking the wallet)
`initializePassword` so we can set a password but skip account onboarding

### DIFF
--- a/.changeset/fair-eagles-serve.md
+++ b/.changeset/fair-eagles-serve.md
@@ -1,0 +1,5 @@
+---
+"@talismn/keyring": patch
+---
+
+feat: added Keyring methods for mobile app compatibility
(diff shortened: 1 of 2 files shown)


alecdwm:
@0xKheops I think that CI failure isn't from anything I did, right?

prury:
@alecdwm this is just paseo testnet not being available, my new PR fixes that

0xKheops (approved):
LGTM!

PR #2407, opened 2026-05-05
feat: proxy accounts

0xKheops:
Adds support for adding/removing proxies for substrate accounts, in the manage accounts page

Misc: 
- refactored substrate submit buttons so they can be disabled arbitrarily by the parent component, for example if we know tx would fail because of insufficient balance. this is a major change as we usually handled this case by displaying a placeholder button instead when a tx was not valid, which was causing flickering.

Worth testing:
- send funds with PV and ledger substrate accounts
- add/remove proxy with PV and ledger substrate accounts, with and without having enough funds

QA:
- Fixes https://github.com/TalismanSociety/talisman-qa/issues/114


prury (dismissed):

prury (dismissed):

0xKheops:
missing password check

prury (dismissed):

PR #2409, opened 2026-05-06
Daily cronjob & flakiness solving

prury:
- Adds assertions to solve flakiness on the accounts test
- cronjob now runs daily


0xKheops (approved):
LGTM

PR #2412, opened 2026-05-07
chore: tidy up packages

alecdwm:
Removes `@polakdot/util` and `@polkadot/util-crypto` from the balances packages, meaning downstream consumers don't need to bloat their apps with their (surprisingly large, since they pull in `@polkadot/wasm-crypto`) contribution to the bundle size.


0xKheops (approved):
♥️

PR #2413, opened 2026-05-08
feat: no pjs packages

alecdwm:
A proof-of-concept done by claude to fully ditch all `@polkadot/` deps from our packages.

Breaks compat with several substrate chains (but NOT polkadot-asset-hub nor bittensor!!) so isn't good to merge, but is a pretty sick PoC if you ask me.

I'm gonna use this in the mobile app for a little while just for fun (no more useless 5MB wasm bundle is nice).

This PR builds ontop of the changes in https://github.com/TalismanSociety/talisman/pull/2396

---

NOTE: This PR has unresolved conflicts with https://github.com/TalismanSociety/talisman/pull/2469 and so it is not currently being used.

After we find papi v2 is suitable to fully replace @polkadot/api for all chains in the extension, we should revive this PR so that we can also exclusively use papi in the mobile app.


alecdwm:
Fixed by https://github.com/TalismanSociety/talisman/pull/2470

PR #2421, opened 2026-05-19
Visual regression test refactor & more routes

prury:
This PR wraps the per-route work in a try/catch so that visual diffs and step errors are reported without breaking the build.

Changes:

 On failure, the error is:
  - Logged to the CI output via console.warn (with a biome-ignore for lint/suspicious/noConsole, since the log is intentional).
  - Attached to the Playwright HTML report as a visual-regression-error annotation for later inspection.

Notes

 - Diff artifacts (-actual.png / -diff.png) are still produced under test-results/, so visual regressions remain reviewable manually.


0xKheops (approved):
LGTM

PR #2426, opened 2026-05-27
Adapt bittensor api tests

prury:
## Changes
 - **`Validators for Targon subnet are returned`**: iterate over every validator (not just the first) and assert`hotkey` matches the SS58 format and `stake > 0`.
 - **New test `Subnets for a validator are returned`**: pulls a live hotkey from Apex (netuid 1) and validates the new `/validators/{hotkey}/subnets` endpoint — checks shape, non-zero stake, unique netuids, and that the source subnet(Apex) is present in the result.

 ## Notes
 - `thirty_day_apy` is intentionally not asserted — the schema allows `null`.
 - No production code changes; tests only.


0xKheops (approved):
LGTM!

Issue #547, opened 2023-02-13
Prevent Sentry clobbering sentry instance in webapp


chidg:
**Describe the bug**
Sentry releases and events are appearing in the dashboard for the Talisman portal webapp.
This may be because the instance of Sentry in the extension is clobbering some window variables on the page when injected.

![image](https://user-images.githubusercontent.com/3326734/218406774-e8f125b2-7c3e-41db-9f9f-e3700b1c4474.png)

chidg:
Closed as solved (no imports of sentry in content script or page.js)

PR #2431, opened 2026-06-04
fix: isolate Sentry from global state

### LINKED ISSUE #2418: Sentry information leaking into other Sentry instances.
This extension is currently using Sentry SDK initialised in the global state: https://github.com/TalismanSociety/talisman/blob/dev/apps/extension/src/sentry.ts#L21

This goes against the recommendation from Sentry to use a manual client when using the SDK in shared environments, such as browser extensions:
- https://docs.sentry.io/platforms/javascript/guides/react/best-practices/shared-environments/

An issue with this configuration has been reported in the past: https://github.com/TalismanSociety/talisman/issues/547

This is a feature request to change the Sentry SDK configuration as per Sentry's recommendation.

0xKheops:
- Fixes https://github.com/TalismanSociety/talisman/issues/2418

Frontend was initialised with SentryReact.init() which registers the client on the global __SENTRY__ carrier and patches global APIs (onerror, fetch, history...), leaking events between the extension and other Sentry SDK instances (#2418, #547).

- share the background's manual BrowserClient + Scope setup (core/config/sentry) with the UI, per https://docs.sentry.io/platforms/javascript/best-practices/shared-environments/
- replace the GlobalHandlers integration with our own window error/unhandledrejection listeners on extension pages
- replace @sentry/react's ErrorBoundary (captures via the global instance) with a custom boundary reporting to the manual client, and drop the @sentry/react dependency
- drop browserTracingIntegration (relies on global state)
- route stray global captureException calls in background code (keyring, migrations, analytics) to the manual client: they were silently dropped since the background never initialises the global instance

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -44,7 +44,6 @@
     "@react-rxjs/core": "^0.10.7",
     "@scure/bip39": "^1.6.0",
     "@sentry/browser": "8.35.0",
-    "@sentry/react": "8.35.0",
     "@sentry/types": "8.35.0",
     "@solana/qr-code-styling": "1.6.0",
     "@solana/spl-token": "^0.4.13",
--- a/apps/extension/src/core/background.ts
+++ b/apps/extension/src/core/background.ts
@@ -18,6 +18,20 @@ import { migrateConnectAllSubstrate } from "./libs/migrations/legacyMigrations"
 
 sentry.init()
 
+// the manual client excludes Sentry's GlobalHandlers integration (it relies on global state),
+// so capture uncaught errors and unhandled rejections with our own listeners.
+// mechanism hints mirror GlobalHandlers so these are reported as unhandled crashes
+self.addEventListener("error", (event) => {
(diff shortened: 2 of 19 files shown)


prury:
Tested a local production build of this branch (release:3.6.3-dev-b1f94e079, Chrome MV3) pointing to the Talisman Sentry project, using jumper.exchange (which ships its own Sentry) as the test dapp.

All tests passed:

  - No leaks in either direction: errors thrown on the dapp page never reached Talisman's Sentry project, and no extension envelopes were sent to the dapp's DSN while connecting the
  wallet, switching networks, and rejecting signatures.
  - Capture still works: unhandled errors/rejections from both the popup and the background service worker arrived in Sentry under the test release. One observation: the background
  error showed mechanism:generic instead of onerror.
  - Privacy setting respected: with error reporting disabled in Settings, no events were sent.
  - Global isolation confirmed: window.__SENTRY__ on extension pages contains only the SDK's debug logger (created at import time) — no globally reachable client/scope.

prury (approved):

PR #2432, opened 2026-06-08
E2E: on-chain transfer verification, derived-accounts & backup coverage

prury:
### On-chain transfer verification (`testnet-transfers.spec.ts`)
The transfer test previously trusted only the wallet UI ("Confirmed"). It now reads the recipient's native balance straight from the local Anvil node (raw `eth_getBalance` JSON-RPC, no new deps) and asserts the balance grew by **exactly** the amount sent. This proves the wallet actually built, signed and broadcast a valid transaction — catching reverted/mis-signed/stuck txs the UI would mask.
- Delta-based assertion → works regardless of the recipient's starting balance.
- `expect.poll` instead of arbitrary waits → handles Anvil's 1s block time reliably.

### Derived accounts + mnemonic backup test (`derived-and-backup.ts`)
New test that imports an account with multiple derived accounts and verifies:
- all derived accounts appear in the sidebar,
- the "recovery phrase(s) not backed up" alert shows in settings,
- the "Protect your funds" backup popup shows on the portfolio.

### Test-infra improvements (`fixtures.ts`)
- `importAccount` now supports a `multipleAccs` option to import N derived accounts.
- Error handling: tests now fail **only** on errors originating from the extension (`chrome-extension://`). Errors thrown by external dapps the test navigates  third-party scripts, etc.) are ignored —
removing a class of unrelated flakiness.


0xKheops (approved):
LGTM!

PR #2488, opened 2026-07-17
⚡️ non-blocking scrypt for json import

0xKheops:
JSON account import runs one scrypt key derivation per keystore (N=2^17, pure JS) synchronously on the UI thread — the page freezes for the whole derivation, once for the file password and once per selected account (sequentially). The existing `setTimeout` wrappers only deferred the freeze, they didn't remove it.

## Changes

- `packages/crypto`: new `decryptPjsKeystoreAsync`, identical to `decryptPjsKeystore` but using `scryptAsync` from @noble/hashes, which yields to the event loop between blocks — same output, UI keeps painting. The keystore parsing/validation is factored into a shared `parseKeystore` so both variants stay byte-compatible. Sync variant untouched (background callers keep it).
- JSON import UI (`pjsImportPairs.ts`): `unlockPair` and `unlockMultiAccountsJson` are now async and use the async decrypt.
- `AccountAddJson/context.ts`: dropped the `setTimeout` wrappers (including the `// hangs UI` one) in `unlockFile` / `unlockAccounts`; the unlock progress bar can now actually animate between per-account derivations.

### DIFF
--- a/.changeset/crypto-async-keystore-decrypt.md
+++ b/.changeset/crypto-async-keystore-decrypt.md
@@ -1,0 +1,5 @@
+---
+"@talismn/crypto": minor
+---
+
+Add `decryptPjsKeystoreAsync` (non-blocking scrypt derivation for UI threads)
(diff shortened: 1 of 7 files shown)


0xKheops:
feels unnecessary

PR #2501, opened 2026-07-18
feat(signing): sr25519 VRF signing via signer.signVrf

nol4lej:
## Motivation

Privacy protocols that build shielded/ZK systems have always faced the same key-management question: where does the user's private identity come from?

Historically there have been only two answers:

1. **A dedicated seed**. Monero, Zcash, Penumbra, Aleo and Railgun all follow this model: the shielded identity has its own root secret, and the user must back up a new seed (or trust a dedicated wallet to custody it). Losing that backup means losing funds.
2. **Deterministic signature derivation**. Aztec v1, Umbra and Fluidkey (ERC-5564) on Ethereum, Elusiv/Light on Solana: the dapp asks the wallet to sign a fixed message and derives the private keys from the signature. No new backup, since the wallet's existing seed covers everything, and the identity is always re-derivable.

The second model is strictly better UX, but it only works where signatures are deterministic: ECDSA (RFC 6979) and ed25519 (RFC 8032). It breaks on **sr25519**, the default curve of every Substrate wallet, because schnorrkel signatures are randomized by design. Every `signRaw` over the same message yields a different signature, so every re-sign silently creates a *different* identity and previously received shielded funds become undecryptable. This is why no Substrate privacy project uses signature derivation today: the ecosystem (e.g. Manta) fell back to model 1, shipping dedicated signers/wallets with their own seed.

The frustrating part is that sr25519 already contains the right primitive: the schnorrkel **VRF**. This is not exotic cryptography bolted on from outside; it is the same VRF Polkadot itself runs on. BABE elects block producers through a per-slot VRF lottery signed with the validator's sr25519 key, and parachain approval assignments rely on it too. That is precisely why `schnorrkel`'s VRF ships in `sp-core`, and why the JS ecosystem already exposes it (`@polkadot/util-crypto`'s `sr25519VrfSign`, `@scure/sr25519`'s `vrf`). Its 32-byte output (`hash_to_point(ctx, msg, pk)^sk`) is fully deterministic for a given (secret key, context, message, extra). Only the accompanying 64-byte proof is randomized, and that proof remains verifiable against the account's public key. Wallets just don't expose it to dapps.

This PR exposes it as `signer.signVrf`, a Talisman-specific extension of the injected API (same pattern as the existing non-spec `encryptMessage`/`decryptMessage`). With it, a dapp can derive a stable private address / ZK identity from an sr25519 account exactly the way Ethereum and Solana dapps already do from theirs: no extra seed, no separate backup, and using a primitive the Polkadot protocol has battle-tested in consensus for years.

Our concrete use case: we're building Orbinum, a Substrate chain with a ZK shielded pool, and this is the missing piece to let sr25519 users derive their private address from the wallet they already have. It brings Substrate accounts into the same signature-derivation flow the wider ZK ecosystem (ERC-5564-style stealth addresses, shielded pools) already standardized on, instead of forcing a dedicated seed like every Substrate privacy project so far.

If you want to see the flow this enables, you can try it today with an EVM wallet at https://app.orbinum.network/wallet/private: connect, sign one fixed message, and a private address for the shielded pool is derived on the spot, fully recoverable from the wallet's own seed. `signer.signVrf` is exactly what's missing to offer that same experience to sr25519 accounts.

## What this adds

New dapp-facing message `pub(vrf.sign)`, wired end to end:

- **`@talismn/crypto`**: `sr25519SignVrf` / `sr25519VerifyVrf`, over `@scure/sr25519`'s `vrf` (already a dependency, no new packages). Returns `output(32) || proof(64)`: the output is deterministic per `(secretKey, origin, context, data)`, only the proof is randomized.
- **The `substrate-vrf` namespace**: the wallet never signs over the caller's raw context. The effective schnorrkel signing context is the frozen frame `"substrate-vrf" || u32_le(origin.len) || utf8(origin) || u32_le(context.len) || context`, where `origin` is the requesting site's web origin (`scheme://host`). This does two things: a dapp cannot use the wallet as a VRF oracle for any other schnorrkel protocol, and every output is bound to the requesting site — no other site can obtain it, not even the same host on another scheme. The frame is wallet-neutral on purpose, so other wallets can adopt it the way `<Bytes>` wrapping is shared for `signRaw`.
- **Injected API**: `signer.signVrf({ address, data, context? })` returns `{ id, signature }` (96 bytes, hex). Feature-detect with `typeof signer.signVrf === "function"`. Dapps verify with `sr25519VerifyVrf(publicKey, data, signature, { origin: location.origin, context })`.
- schnorrkel's `extra` is deliberately not exposed: it changes the proof but never the 32-byte output, so a dapp using it as a domain separator would silently derive one identity where it expects several. A payload carrying `extra` is rejected rather than ignored.
- **Background**: a new `vrf-sign` request type through the standard request store, with the same site-authorization checks as `pub(bytes.sign)`. Payload fields are strictly validated (even-length `0x` hex only — `hexToU8a` silently rewrites malformed input — with size bounds) both at the request boundary and again at approval. Requests are rejected up front for anything but local sr25519 keypair accounts (hardware, vault, watch-only and ethereum accounts can't VRF-sign), and the approval handler re-asserts the curve before touching the secret key.
- **UI**: a dedicated approval popup (`vrf-sign/:id`) showing the signed data and the caller context, plus inclusion in the pending-request badge count.

The copied polkadot-js wire types in `pjsInterop.ts` are untouched. The new message lives in Talisman's own `SigningMessages` domain map, so the legacy protocol surface stays verbatim.

## Testing

- **Unit (`packages/crypto`)**: schnorrkel parity vectors generated with `@polkadot/util-crypto` 14.0.3 (wasm) for the underlying primitives; byte-exact assertions on the frozen frame layout; origin and context domain separation; no verification across the namespace boundary in either direction.
- **Integration (`Extension.spec.ts`)**: full background flow — deterministic output across approvals, verification through the namespace (including that another origin, another scheme on the same host, or the bare host cannot verify), malformed/oversized/`extra` payload rejection, account-type gating, and refusal to answer a request queued under another signing type.
- **E2E (Playwright)**: self-contained local dapp, connect popup, two `signVrf` approvals through the real popup: identical output, distinct proofs, context displayed and separating outputs.

## Notes for reviewers

- `context` is optional hex; omitted means empty, `"0x"` is an explicit empty. polkadot-js `sr25519VrfSign` byte-parity holds for the underlying primitives (covered by the unit vectors) but not through `signer.signVrf`, which always signs inside the namespaced frame.
- The response is signature-shaped (`{ id, signature }`) to stay consistent with `signRaw`.


0xKheops:
Thanks for this PR. I've pushed a series of commits on top of your branch that reshape the API before it freezes: VRF outputs are deterministic, so everything about the signing context has to be right in the first release. Any later change would rotate every derived identity. 

Summary of what changed and why:

**The signing context is now namespaced and origin-bound.** The wallet never signs over the caller's raw `context`. The effective schnorrkel context is a fixed frame — `"substrate-vrf" || u32_le(origin.len) || utf8(origin) || u32_le(context.len) || context` — where `origin` is the requesting site's web origin (`scheme://host`). Two reasons: a dapp can't use the wallet to produce VRF outputs for another schnorrkel protocol's transcript, and outputs are per-site — the `(data, context)` pair is public in a dapp's JS, so without origin binding any other site authorized for the account could request the same output. The tag is deliberately wallet-neutral so other wallets can adopt the same frame, the way `<Bytes>` wrapping is shared for `signRaw`. The layout is asserted byte-exactly in tests and documented as frozen; a future revision would be a new opt-in tag, never a replacement.

**`extra` was removed from the API.** schnorrkel feeds `extra` into the proof transcript only — it never changes the 32-byte output. A dapp using it as a domain separator would silently derive one identity where it expects several (e.g. spending key == viewing key). A payload carrying `extra` is now rejected rather than ignored, so the misunderstanding surfaces immediately. Re-adding it later is additive, since outputs never depended on it.

**Validation and request handling.** Payload hex is strictly validated (even-length `0x`-prefixed, size-bounded) at the request boundary and re-validated at approval — `hexToU8a` silently rewrites malformed input, which would let the bytes the user approved and the bytes signed diverge. All approval paths now assert the request type (they share an id space, and a `SignerPayloadRaw` also parses as a VRF payload), and a failed approval rejects the queued request instead of leaving it pending.

**Crypto surface.** The public exports are `sr25519SignVrf` / `sr25519VerifyVrf`, which take `{ origin, context }`. The raw polkadot-js-parity primitives stay internal to the package, covered by wasm-schnorrkel parity vectors in the unit tests.

**UI.** The approval popup shows the full transcript (data + context) with the standard sign-request wording.

For a dapp integrating this, two things differ from the original API in this PR: `extra` is gone, and verification is now `sr25519VerifyVrf(publicKey, data, signature, { origin: location.origin, context })` from `@talismn/crypto`. A dapp served from several origins derives a different value on each, so pin a canonical one. The "What this adds" / "Notes for reviewers" sections of the description no longer match on those points (`extra`, the `context: "0x73…"` pjs-parity recipe) — would you mind updating them? [edit: i was able to update description myself]

nol4lej:
Thank you for the thorough review and for taking this implementation the rest of the way, the reshaping you did goes well beyond what I'd have caught on my own, and doing it before the freeze is exactly right.

Integrated and tested end to end against a local build of this branch, deterministic, and origin binding confirmed with a real sr25519 account. Agreed on all four points.

On `extra`: we'd exposed it in the original PR mirroring `vrf_sign_extra` argument-for-argument, and my docstring even claimed the output depended on it, which was wrong. Confirmed against `@scure/sr25519`, two different `extra` values give byte-identical outputs. Rejecting the payload outright is the right call; ignoring it would leave the footgun silent.

On origin binding: this closes the one thing we couldn't fix dapp-side. We derive a spending key from the output, so before this any authorized origin could reconstruct a user's private identity by requesting the same `(data, context)`. Our own domain separator was never going to fix that, only the wallet can supply an origin the page can't forge. Making the tag wallet-neutral so other wallets can adopt the same frame is a nice touch.

We verify locally with `sr25519VerifyVrf`before the bytes seed anything, so a wallet that ignored the namespace fails hard instead of silently deriving an unbound identity.

PR #2561, opened 2026-09-11
♻️ migrate translation extraction to i18next-cli

0xKheops:
Replace deprecated `i18next-parser` with `i18next-cli`, preserving the SimpleLocalize upload workflow.

- Share runtime/extraction settings, support `<Trans defaults>`, and fix dynamic interpolation keys.
- Bundle 19 new/corrected translations across all 12 languages; exclude developer READMEs from extraction.
- Require Node 22+ and add extraction, catalog, and browser regression tests.

Validated: CI passed, including unit/e2e tests and translation upload. Browser checks cover phishing, contact deletion, and network/token reset dialogs in English and French.

Manual QA (pending), using the PR build in English and another language:

- [ ] Inspect swap recipient warnings, dapp add-network requests, and Polkadot Vault/QR setup.
- [ ] Inspect staking confirmations with 0% and nonzero auto-compound.
- [ ] Check names, symbols, amounts, percentages, links, and wrapping; no raw placeholders or unexpected English fallback. Cancel before submission; transactions are unnecessary.


playwright/e2e-tests/i18next-migration.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    await page.goto(`${dashboard}/settings/general/language`)
+    await page.getByRole("button", { name: label, exact: true }).click()
+    await expect(page.locator("html")).toHaveAttribute("lang", language)
+
+    await page.goto(`${dashboard}/phishing-page-detected/${encodeURIComponent(phishingUrl)}`)
+    await expect(
+      page.getByText(plain(phishingKey, { displayUrl: phishingUrl }), { exact: true })
+    ).toBeVisible()
+    await expect(page.getByText(phishingUrl, { exact: true })).toBeVisible()
+    await page.screenshot({ path: testInfo.outputPath(`phishing-${language}.png`) })
```
prury: page.screenshot() only writes an artifact, nothing is compared, so this can't fail. We already have visual-regression.spec.ts using toHaveScreenshot(name, { maxDiffPixelRatio: 0.01 }).

playwright/e2e-tests/i18next-migration.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  ["fr", "Français"],
+]) {
+  test(`translated warnings and confirmations (${language})`, async ({
+    onboardedPage: page,
+    extensionId,
+  }, testInfo) => {
+    const catalog: Record<string, string> = JSON.parse(
+      readFileSync(path.resolve("apps/extension/public/locales", language, "common.json"), "utf8")
+    )
+    const text = (key: string) => catalog[key] || key
```
prury: The || key fallback makes most of the assertions below unable to fail.

playwright/e2e-tests/i18next-migration.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    await page.goto(`${dashboard}/settings/address-book`)
+    await page.getByRole("button", { name: text("Add a contact"), exact: true }).click()
+    await page.getByPlaceholder(text("Contact name"), { exact: true }).fill(contactName)
+    await page
+      .getByPlaceholder(text("Address"), { exact: true })
+      .fill("0x0000000000000000000000000000000000000001")
+    await page.getByRole("button", { name: text("Save"), exact: true }).click()
+    const contact = page.getByText(contactName, { exact: true })
+    await expect(contact).toBeVisible()
+    await contact.locator("../..").getByRole("button").last().click()
```
prury: Two levels of DOM traversal plus "last button" will break on any markup change. The convention in this repo is test ids: 90 data-testid in the source and 57 getByTestId across the other specs (17 in testnet-transfers.spec.ts alone). This file uses none.  Same issue at 65 (getByRole("textbox").first()) and 89

playwright/e2e-tests/i18next-migration.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    await page.getByRole("button", { name: text("Delete contact"), exact: true }).click()
+    await expect(page.getByText(plain(deleteKey, { contactName }), { exact: true })).toBeVisible()
+    await expect(page.getByText(contactName, { exact: true }).last()).toBeVisible()
+    await expect(page.locator("body")).not.toContainText("{{")
```
prury: This is the one assertion that catches the actual regression this migration risks: a <Trans> whose values don't line up with the {{placeholders}} renders them literally.

Right now it only runs on the contact modal. Could we move it to an afterEach, or repeat it after each navigation? The phishing page, network reset and token reset are all rewritten <Trans> blocks and none of them get this check.

playwright/e2e-tests/i18next-migration.spec.ts
0xKheops: Agreed. Removed the screenshots in 2c6e36c9d; the text assertions are the real check. Baselines for toHaveScreenshot would need eight new linux-only snapshots plus a change to the update-snapshots workflow, not worth it here.

playwright/e2e-tests/i18next-migration.spec.ts
0xKheops: Agreed. The lookup now throws on a missing or empty catalog entry (2c6e36c9d), so a key dropped or reshaped by extraction fails the test instead of matching the key literal.

playwright/e2e-tests/i18next-migration.spec.ts
0xKheops: Agreed. Now uses getByTestId("contact-menu-button") (added to the row trigger), input[name="name"] on the network form (added the name attribute, the token form already had one) and input[name="symbol"]. 2c6e36c9d.

playwright/e2e-tests/i18next-migration.spec.ts
0xKheops: Added an expectNoRawPlaceholders helper called after each of the four views (2c6e36c9d). afterEach would run after the modals close, so per-view is the right spot. Small nuance: the exact-text assertions built from the catalog with substituted values already fail on a literal {{placeholder}} for each of those four Trans blocks; the body check is a wider net for any other Trans on the page.

prury (approved):

